PR #2, opened 2016-08-10
Adds initial reporters: urlconnection and kafka08

adriancole:
This adds a couple reporters that have been copy-pasted at least thrice.

For starters, this re-uses the existing interface AsyncSpanConsumer, as
opposed to defining a new one called Reporter.


adriancole:
@kristofa @michaelsembwever @marcingrzejszczak @nicmunroe lemme know if this code would be helpful to you if pulled out into a separate codebase like this.

kafka08/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.zipkin.reporter</groupId>
+    <artifactId>zipkin-reporter-parent</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>zipkin-reporter-kafka08</artifactId>
```
adriancole: note: naming the artifact and java package kafka08 to avoid repeating sins of the past. We'll eventually need to support kafka 10!

README.md
```diff
@@ -1,2 +1,30 @@
...
-Shared library for reporting zipkin spans on transports such as http or kafka
+[![Gitter chat](http://img.shields.io/badge/gitter-join%20chat%20%E2%86%92-brightgreen.svg)](https://gitter.im/openzipkin/zipkin) [![Build Status](https://travis-ci.org/openzipkin/zipkin-java-reporter.svg?branch=master)](https://travis-ci.org/openzipkin/zipkin-java-reporter) [![Download](https://api.bintray.com/packages/openzipkin/maven/zipkin-java-reporter/images/download.svg) ](https://bintray.com/openzipkin/maven/zipkin-java-reporter/_latestVersion)
+
+# zipkin-java-reporter
+Shared library for reporting zipkin spans on transports including http and kafka.
+
+# Usage
+These components can be called when spans have been recorded and ready to send to zipkin.
+
+For example, you may have a class called Recorder, which flushes on an interval. The reporter
```
adriancole: here's the usage example

urlconnection/src/main/java/zipkin/reporter/urlconnection/URLConnectionReporter.java
```diff
@@ -0,0 +1,127 @@
...
+        callback.onError(e);
+        if (e instanceof Error) throw (Error) e;
+      }
+    });
+  }
+
+  void send(List<Span> spans) throws IOException {
+    byte[] body = Codec.JSON.writeSpans(spans);
+    String mediaType = "application/json";
+    URL url = new URL(baseUrl() + (baseUrl().endsWith("/") ? "" : "/") + "api/v1/spans");
```
nicmunroe: Is there a use case where the baseUrl could change during runtime? Why not create the URL once during app construction and reuse it?

nicmunroe:
It would be great to have the reporters pulled into separate modules like this.

urlconnection/src/main/java/zipkin/reporter/urlconnection/URLConnectionReporter.java
adriancole: Good point. Will make it possible to affect dynamically.

adriancole: done

urlconnection/src/main/java/zipkin/reporter/urlconnection/URLConnectionReporter.java
```diff
@@ -0,0 +1,154 @@
...
+public abstract class URLConnectionReporter implements AsyncSpanConsumer {
+
+  /**
+   * Returns the POST URL for zipkin's <a href="http://zipkin.io/zipkin-api/#/">v1 api</a>, usually
+   * "http://zipkinhost:9411/api/v1/spans".
+   *
+   * <p>This is exposed for those who wish to set this dynamically.
+   */
+  // @FunctionalInterface except library compatible w/ JRE 7
+  public interface PostURL {
```
adriancole: @nicmunroe here's the dynamic thing. This accepts string so that the caller doesn't need to deal with URL's nastiness

nicmunroe: I was thinking more that it was wasteful to incur the cost of recreating the same URL for every request if there was no need for it to be dynamic. But thinking about it some more: (1) this only happens once on a period basis - e.g. once per second, and (2) it's cheap to create a URL anyway. So having the possibility for a dynamic URL doesn't really cost much other than some minor code maintenance hassle. 👍

adriancole: Thanks for clueing me into your thought process. I like how you think! On
that note, will add benchmarks at some point in the near future.

PR #3, opened 2016-08-11
Adds Reporter interface

adriancole:
This adds a Reporter interface, which makes more sense than re-using the
SpanConsumer one, if no reason except the javadoc can be more helpful.

### DIFF
--- a/core/src/main/java/zipkin/reporter/internal/AwaitableCallback.java
+++ b/core/src/main/java/zipkin/reporter/internal/AwaitableCallback.java
@@ -1,0 +1,66 @@
+/**
+ * Copyright 2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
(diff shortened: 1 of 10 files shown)


adriancole:
cc @nicmunroe

core/src/main/java/zipkin/reporter/internal/AwaitableCallback.java
```diff
@@ -0,0 +1,65 @@
...
+    try {
+      while (true) {
+        try {
+          countDown.await();
+          Object result = throwable.get();
+          if (result == null) return;
+          if (result instanceof Throwable) {
+            if (result instanceof Error) throw (Error) result;
+            if (result instanceof RuntimeException) throw (RuntimeException) result;
+            throw new RuntimeException((Exception) result);
```
schlosna: if `result` is an instance of `InterruptedException`, do you want to set `interrupted = true` before wrapping and rethrowing, otherwise you may lose that the error was due to an interruption?

nicmunroe: Recommend `throw new RuntimeException((Throwable)result)` to prevent class cast exceptions in case something bizarre was thrown - nothing prevents someone from creating a custom class that extends `Throwable`.

adriancole: you've seen evils I hope I'll never see :)

adriancole: @schlosna good point.. we're pretending that everything happened on the thread calling await, so if result is an InterruptedException (which happened on a callback thread), we should pretend it happened here (like you said)

adriancole:
thx for the reviews, folks. fixed

PR #4, opened 2016-08-14
Unimplements dynamic http urls as it wasn't actually requested

adriancole:
I misunderstood a request from @nicmunroe about caching the URL and then
    pre-maturely implemented dynamic urls. In practice we've not had a
    request for dynamic URLs with URLConnection.
    
    Long story short: this undoes the fanciness as better sooner than later.


adriancole:
cc @nicmunroe

nicmunroe:
LGTM! 👍

Issue #1, opened 2016-08-10
Create shared libraries for reporting zipkin spans on the JVM


adriancole:
_From @adriancole on April 30, 2016 0:50_

Those writing instrumentation on the JVM have fairly undifferentiated tasks, putting spans on a transport for collection. For example, code in htrace, brave and spring-cloud-sleuth are nearly identical, except the recording step (taking a library-specific type into a zipkin representation). Most of these projects lag in transport diversity because they have to individually write support for kafka, etc.

By adding a Reporter to zipkin-java, those who wish to share a library (or repackage it internally), can opt-into a tested, dependency-light plugin to report data to zipkin, which they don't have to maintain.

Some possible outcomes include..
- A load generator tool, which can compare performance using any of the zipkin transports #45
- a generic reporter which makes it possible for finagle to send to _any_ zipkin transport, instead of just [scribe](https://github.com/twitter/finagle/issues/465) 
- reducing maintenance in brave by removing the [3 span collectors](https://github.com/openzipkin/brave) (reporters) in favor of a generic one 
- increase diversity of spring-cloud-sleuth beyond just reporting to [http](https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporter.java)
- replace or augment the [htrace zipkin transport](https://github.com/apache/incubator-htrace/blob/master/htrace-zipkin/src/main/java/org/apache/htrace/Transport.java), so that it reaches beyond scribe and kafka and shares tests

_Copied from original issue: openzipkin/zipkin-java#181_

adriancole:
cc @devinsba @kristofa @eirslett @michaelsembwever @marcingrzejszczak @clehene @mansu @klingerf @CodingFabian

adriancole:
_From @clehene on May 8, 2016 18:45_

This could be a good way to provide some useful cross library reusability.

One of my goals with the htrace zipkin transport was to push it up and provide a general transport for htrace, so I kept it decoupled from the actual htrace - zipkin translation logic (note [here](https://github.com/apache/incubator-htrace/blob/master/htrace-zipkin/src/main/java/org/apache/htrace/impl/KafkaTransport.java#L70) that the list of spans is just a list of byte arrays. 

I think it would be worth to maintain this decoupling so that the transporters ;) could be reused for other trace payloads. 
[...]

PR #6, opened 2016-08-20
Renames Reporter to Sender, which accepts a list of encoded spans

adriancole:
Sender is a component which puts encoded spans onto the transport. As
discussed in #1, there are use cases where other libraries or systems
may be in charge of marshalling spans.

The Sender interface includes the ability to do a health check. This is
necessary for fail-fast scenarios, such as when Kafka is down.

The Reporter interface will return in a future pull request.

Here's a mythical version of how this might be used:

```java
class Reporter implements Flushable {

  --snip--
  URLConnectionSender sender = URLConnectionSender.builder()
                                                  .endpoint("http://localhost:9411/api/v1/spans")
                                                  .build();

  Callback callback = new IncrementSpanMetricsCallback(metrics);

  // Is the connection healthy?
  public boolean ok() {
    return sender.check().ok;
  }

  public void report(Span span) {
    pending.add(SpanEncoder.THRIFT.encode(span));
  }

  @Override
  public void flush() {
    if (pending.isEmpty()) return;
    List<byte[]> drained = new ArrayList<byte[]>(pending.size());
    pending.drainTo(drained);
    if (drained.isEmpty()) return;

    sender.sendSpans(drained, callback);
  }
```


adriancole:
@nicmunroe this is the change I wanted you to see..

adriancole:
adjusted design to pretty much exactly what htrace uses, as for example the logging use case (scribe) requires custom list encoding. For example, scribe encodes each span as a base64 encoded log line.

https://github.com/apache/incubator-htrace/blob/master/htrace-zipkin/src/main/java/org/apache/htrace/Transport.java

adriancole:
Here are the benchmark results. Note that creating a more performant encoder is easier now that we only need to implement one method (`SpanEncoder.encode(Span)`) vs the whole zipkin Codec interface.

```
Benchmark                                        Mode  Cnt    Score    Error   Units
EncodingBenchmarks.encode100Spans_json          thrpt   15    0.625 ±  0.035  ops/ms
EncodingBenchmarks.encode100Spans_thrift        thrpt   15   10.511 ±  0.428  ops/ms
EncodingBenchmarks.encodeListOf100Spans_json    thrpt   15  133.351 ±  1.432  ops/ms
EncodingBenchmarks.encodeListOf100Spans_thrift  thrpt   15  261.234 ± 14.926  ops/ms
```

PR #7, opened 2016-08-21
Adds Reporter for instrumentation to use after they record a span

adriancole:
## Reporter
After recording an operation into a span, it needs to be reported out of process. There are two
builtin reporter implementations in this library, although you are free to create your own.

The simplest mechanism is printing out spans as they are reported.

```java
Reporter.CONSOLE.report(span);
```

Reporter is very similar to the following:
* Finagle's DeadlineSpanMap
* zipkin-finagle's SpanRecorder
* Brave's SpanCollector
* HTrace's ZipkinSpanReceiver
* Wingtips' ZipkinSpanSender
* Sleuth's ZipkinSpanReporter
* Jaeger's Reporter

and.. probably many others.

## AsyncReporter
AsyncReporter is how you actually get spans to zipkin. By default, it waits up to a second
before flushes any pending spans out of process via a Sender.

```java
reporter = AsyncReporter.builder(URLConnectionSender.create("http://localhost:9411/api/v1/spans"))
                        .build(Encoder.THRIFT_BYTES);

// Schedules the span to be sent, and won't block the calling thread on I/O
reporter.report(span);
```

### DIFF
--- a/core/src/main/java/zipkin/reporter/AsyncReporter.java
+++ b/core/src/main/java/zipkin/reporter/AsyncReporter.java
@@ -1,0 +1,261 @@
+/**
+ * Copyright 2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.reporter;
+
+import java.io.Flushable;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Logger;
+import zipkin.Component;
+
+import static java.lang.String.format;
+import static java.util.logging.Level.WARNING;
+import static zipkin.internal.Util.checkArgument;
+import static zipkin.internal.Util.checkNotNull;
+
+/**
+ * As spans are reported, they are encoded and added to a pending queue. The task of sending spans
+ * happens on a separate thread which calls {@link #flush()}. By doing so, callers are protected
+ * from latency or exceptions possible when exporting spans out of process.
+ *
+ * <p>Spans are bundled into messages based on size in bytes or a timeout, whichever happens first.
+ *
+ * @param <S> type of the span, usually {@link zipkin.Span}
+ */
+public abstract class AsyncReporter<S> implements Reporter<S>, Flushable, Component {
+  /**
+   * After a certain threshold, spans are drained and {@link Sender#sendSpans(List, Callback) sent}
+   * to Zipkin collectors.
+   */
+  public static <B> Builder<B> builder(Sender<B> sender) {
+    return new Builder<>(sender);
+  }
+
+  /**
+   * Calling this will flush any pending spans to the transport on the current thread.
+   *
+   * <p>Note: If you set {@link Builder#messageTimeout(long, TimeUnit) message timeout} to zero, you
+   * must call this externally as otherwise spans will never be sent.
+   *
+   * @throws IllegalStateException if closed
+   */
+  @Override public abstract void flush();
+
+  /** Shuts down the sender thread, and increments drop metrics if there were any unsent spans. */
+  @Override public abstract void close();
+
+  public static final class Builder<B> {
+    final Sender<B> sender;
+    ReporterMetrics metrics = ReporterMetrics.NOOP_METRICS;
+    int messageMaxBytes;
+    long messageTimeoutNanos = TimeUnit.SECONDS.toNanos(1);
+    int queuedMaxSpans = 10000;
+    int queuedMaxBytes = onePercentOfMemory();
+
+    static int onePercentOfMemory() {
+      long result = (long) (Runtime.getRuntime().totalMemory() * 0.01);
+      // don't overflow in the rare case 1% of memory is larger than 2 GiB!
+      return (int) Math.max(Math.min(Integer.MAX_VALUE, result), Integer.MIN_VALUE);
+    }
+
+    Builder(Sender<B> sender) {
+      this.sender = checkNotNull(sender, "sender");
+      this.messageMaxBytes = sender.messageMaxBytes();
+    }
+
+    /**
+     * Aggregates and reports reporter metrics to a monitoring system. Should be {@link
+     * ReporterMetrics#forTransport(String) scoped to this transport}.  Defaults to no-op.
+     */
+    public Builder<B> metrics(ReporterMetrics metrics) {
+      this.metrics = checkNotNull(metrics, "metrics");
+      return this;
+    }
+
+    /**
+     * Maximum bytes sendable per message including {@link Sender.MessageEncoding#overheadInBytes(int)
+     * overhead}. Defaults to, and is limited by {@link Sender#messageMaxBytes()}.
+     */
+    public Builder messageMaxBytes(int messageMaxBytes) {
+      checkArgument(messageMaxBytes >= 0, "messageMaxBytes < 0: %s", messageMaxBytes);
+      this.messageMaxBytes = Math.min(messageMaxBytes, sender.messageMaxBytes());
+      return this;
+    }
+
+    /**
+     * Default 1 second. 0 implies spans are {@link #flush() flushed} externally.
+     *
+     * <p>Instead of sending one message at a time, spans are bundled into messages, up to {@link
+     * Sender#messageMaxBytes()}. This timeout ensures that spans are not stuck in an incomplete
+     * message.
+     *
+     * <p>Note: this timeout starts when the first unsent span is reported.
+     */
+    public Builder messageTimeout(long timeout, TimeUnit unit) {
+      checkArgument(timeout >= 0, "timeout < 0: %s", timeout);
+      this.messageTimeoutNanos = unit.toNanos(checkNotNull(timeout, "timeout"));
+      return this;
+    }
+
+    /** Maximum backlog of spans reported vs sent. Default 10000 */
+    public Builder queuedMaxSpans(int queuedMaxSpans) {
+      this.queuedMaxSpans = queuedMaxSpans;
+      return this;
+    }
+
+    /** Maximum backlog of span bytes reported vs sent. Default 1% of heap */
+    public Builder queuedMaxBytes(int queuedMaxBytes) {
+      this.queuedMaxBytes = queuedMaxBytes;
+      return this;
+    }
+
+    /** Builds an async reporter that encodes spans as they are reported. */
+    public <S> AsyncReporter<S> build(Encoder<S, B> encoder) {
+      checkNotNull(encoder, "encoder");
+      checkArgument(encoder.encoding() == sender.messageEncoding().encoding(),
+          "Sender.encoding() %s != Encoder.encoding() %s", sender.messageEncoding().encoding(),
+          encoder.encoding());
+
+      final BoundedAsyncReporter<S, B> result = new BoundedAsyncReporter<>(this, encoder);
+
+      if (messageTimeoutNanos > 0) { // Start a thread that flushes the queue in a loop.
+        final BufferNextMessage<B> consumer =
+            new BufferNextMessage<>(sender.messageEncoding(), messageMaxBytes, messageTimeoutNanos);
+        new Thread(() -> {
+          try {
+            while (!result.closed.get()) {
+              result.flush(consumer);
+            }
+          } finally {
+            for (B next : consumer.drain()) {
+              result.pending.offer(next, result.encoder.sizeInBytes(next));
+            }
+            result.close.countDown();
+          }
+        }, "AsyncReporter(" + sender + ")").start();
+      }
+      return result;
+    }
+  }
+
+  static final class BoundedAsyncReporter<S, B> extends AsyncReporter<S> {
+    static final Logger logger = Logger.getLogger(BoundedAsyncReporter.class.getName());
+    final AtomicBoolean closed = new AtomicBoolean(false);
+    final Encoder<S, B> encoder;
+    final ByteBoundedQueue<B> pending;
+    final Sender<B> sender;
+    final int messageMaxBytes;
+    final int messageOverheadOfSingleton;
+    final long messageTimeoutNanos;
+    final CountDownLatch close;
+    final ReporterMetrics metrics;
+
+    BoundedAsyncReporter(Builder builder, Encoder<S, B> encoder) {
+      this.pending = new ByteBoundedQueue(builder.queuedMaxSpans, builder.queuedMaxBytes);
+      this.sender = builder.sender;
+      this.messageMaxBytes = builder.messageMaxBytes;
+      this.messageOverheadOfSingleton = sender.messageEncoding().overheadInBytes(1);
+      this.messageTimeoutNanos = builder.messageTimeoutNanos;
+      this.close = new CountDownLatch(builder.messageTimeoutNanos > 0 ? 1 : 0);
+      this.metrics = builder.metrics;
+      this.encoder = encoder;
+    }
+
+    /** Returns true if the was encoded and accepted onto the queue. */
+    @Override
+    public void report(S span) {
+      checkNotNull(span, "span");
+      metrics.incrementSpans(1);
+      B next = encoder.encode(span);
+      int nextSizeInBytes = encoder.sizeInBytes(next);
+      metrics.incrementSpanBytes(nextSizeInBytes);
+      if (closed.get() ||
+          // don't enqueue something larger than we can drain
+          messageOverheadOfSingleton + nextSizeInBytes > messageMaxBytes ||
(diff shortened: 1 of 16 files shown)


adriancole:
need to make tests, etc (mostly port them from other places).

@nicmunroe lemme know what you think

adriancole:
@schlosna ik that you're on holiday, but probably you'll be interested in this. it is pre-1.0 so we can adjust as we learn

adriancole:
also cc @felixbarny

adriancole:
going to rejig batchreporter a bit so that pooled buffers (like netty or okio) are possible.

core/src/main/java/zipkin/reporter/BatchReporter.java
```diff
@@ -0,0 +1,209 @@
...
+    this.metrics = builder.metrics == null ? ReporterMetrics.NOOP_METRICS : builder.metrics;
+    this.pending = new LinkedBlockingDeque<>(builder.maxQueueDepth);
+    this.flusher = builder.flushInterval > 0 ? new Flusher(this, builder.flushInterval,
+        "Flusher(" + sender + ")") : null;
+  }
+
+  @Override
+  public void report(Span span) {
+    metrics.incrementSpans(1);
+    if (!pending.offer(spanEncoder.encode(span))) {
```
nicmunroe: Any thought to alternate behavior besides simply dropping the span? You could have it immediately flush the sender with whatever you have even if it's not the regularly scheduled sending time. Maybe the ability to have it auto-flush like that when it's 80% full or something?

These alternate behaviors could add some non-trivial complexity, so maybe it's not worth it at this point (or ever depending on if people would actually use it). Something to think about though.

nicmunroe:
@adriancole Looks good to me. Separating out the reporter vs. the sender makes a lot of sense and is the direction I was accidentally heading.

core/src/main/java/zipkin/reporter/BatchReporter.java
adriancole: Good thoughts. So I'd expect the queue (or buffer) to always be bounded.
What do do when it is full (besides dropping) is a good question. There's
another matter which is whether to drop a span too large to send. For
example kafka has a max message size which people have (usually
erroneously) exceeded.

You are right that we can rejig things to allow a flush ahead of schedule
(based on size) though that would more of a precaution than a cure of a
full queue.

Lets revisit this again on next round.. thanks for the notes!

adriancole:
Added rough-cut of byte tracking and the ability to support non-bytearray buffers

adriancole:
before merging this, I'm going to practice with it by adding tests that bind it to at least htrace and brave. That will show how it works in real life.

adriancole:
just pushed a boatload of integration examples. I'm not currently happy with the api, just using the examples to help me drive down complexity (especially around building things which may or may not need to be transformed)

adriancole:
alright, so not happy with the api, yet, but getting there. obv need to backfill tests. Here are some notes:

* added BoundedQueue, which allows you to drain up to a certain size in bytes. It guards entry to ensure you can't put something larger than that on. This can be used to ask for up to N bytes to go to the sender at a time (as opposed to N spans). This only works when you can size the object (fine in json or thrift encoding).

* changed to htrace's draining approach. Instead of using a scheduler, this polls until there are spans available (up to 1 second as opposed to before which was minimum 1 second). This also controls the shutdown frequency. The difference now is that we report as soon as spans arrive (or as long as there's a backlog). we could pretty easily put a throttle on it later, if need be.

cc @nicmunroe don't pay too much attention to the details of the code, as I've another day of work cleaning it up. That said, lemme know if the approach and the api integration looks alright.

core/src/main/java/zipkin/reporter/AsyncReporter.java
```diff
@@ -0,0 +1,253 @@
...
+      );
+      new Thread(() -> {
+        try {
+          while (!closed.get()) {
+            result.flush();
+          }
+        } finally {
+          shutdown.countDown();
+        }
+      }, "Flusher(" + builder.sender + ")").start();
```
adriancole: I know daemon threads are a polarizing topic.. if we make this daemon, then people could forget to call close and process shutdown will silently toss things.. If we don't make this daemon, someone who forgets to call close will have their process hung and even looking at this thread name.. maybe they will still complain.

Any thoughts?

core/src/main/java/zipkin/reporter/internal/BoundedQueue.java
```diff
@@ -0,0 +1,150 @@
...
+import com.google.auto.value.AutoValue;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
+import zipkin.internal.Nullable;
+
+/** Special queue that can measure the size of entries as they are added. */
+public final class BoundedQueue<E> {
```
adriancole: loosely based on ArrayBlockingQueue

core/src/main/java/zipkin/reporter/internal/BoundedQueue.java
```diff
@@ -0,0 +1,150 @@
...
+    this.byteCounter = byteCounter;
+    this.maxByteCount = maxByteCount;
+  }
+
+  /** Returns true if the element could be added or false if the queue is full */
+  public boolean offer(E element) {
+    Object toAdd = element;
+    if (byteCounter != null) {
+      int byteCount = byteCounter.byteCount(element);
+      if (byteCount > maxByteCount) return false;
```
adriancole: returning false, here, but I could make this a numeric return like caffeine has. we could define one for full and another for over limit. throwing exception is possible, too, though probably not as cheap as a number.

Ex. here's caffeine's codes..
```java
interface Buffer<E> {
  int FULL = 1;
  int SUCCESS = 0;
  int FAILED = -1;
```

https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Buffer.java#L35

reporter/wingtips/src/main/java/zipkin/reporter/wingtips/WingtipsSpanReporter.java
```diff
@@ -0,0 +1,51 @@
...
+import zipkin.reporter.Encoder;
+import zipkin.reporter.Reporter;
+import zipkin.reporter.Sender;
+
+/** Example bridge between Wingtips' ZipkinSpanSender and this project's reporter */
+public class WingtipsSpanReporter implements ZipkinSpanSender, Reporter<Span>, Closeable {
+  final AsyncReporter<Span> delegate;
+
+  public WingtipsSpanReporter(Sender<byte[]> sender) {
+    this.delegate = AsyncReporter.builder(sender).build(Encoder.THRIFT_BYTES);
```
adriancole: doing sender as the first arg and encoder as a terminal arg helps a bit as it keeps the builder less complex. Not sure if this pattern is objectional or not..

adriancole: one limitation is it prevents the "toBuilder()" pattern from working when we do it like this... I'm not terribly concerned about this as I don't think people will be boilerplating async reporters.

adriancole:
ps not obvious yet, but the components in this PR (particularly bounded queue) support a policy something like this, which can be enforced very early (ex as soon as the span is reported), vs now (very late.. like all the way at kafka!)

maxBufferedSpanBytes = 32MiB // currently unenforcable without this code
maxSpansBytesPerMessage = 1MiB // currently transport-specific

core/src/main/java/zipkin/reporter/internal/BoundedQueue.java
```diff
@@ -0,0 +1,150 @@
...
+    this.maxBatchSize = maxBatchSize;
+    this.byteCounter = byteCounter;
+    this.maxByteCount = maxByteCount;
+  }
+
+  /** Returns true if the element could be added or false if the queue is full */
+  public boolean offer(E element) {
+    Object toAdd = element;
+    if (byteCounter != null) {
+      int byteCount = byteCounter.byteCount(element);
```
felixbarny: Does that mean spans have to be serialized twice? Once to get the size and once more to actually send them over the wire?

core/src/main/java/zipkin/reporter/internal/BoundedQueue.java
felixbarny: I'd prefer enums over int constants any time.

core/src/main/java/zipkin/reporter/internal/BoundedQueue.java
adriancole: nope. this just reports the size of the encoded buffer. ex in the case of a
byte array bytes.length

felixbarny: But as far as I understand, `E element` is an instance of `Span` and not a `byte[]`.

core/src/main/java/zipkin/reporter/internal/BoundedQueue.java
adriancole: sgtm

core/src/main/java/zipkin/reporter/internal/BoundedQueue.java
felixbarny: Ok, I see `E` is a `byte[]`: `URLConnectionSender<byte[]> create(String endpoint)`

adriancole: So this type requires a lambda `ByteCounter` to be passed, to count the size of `E` (there's no notion of span in this type). We bind that to `Encoder::byteCount`, which usually binds `E` to `byte[]`. (ex as `Encoder.THRIFT_BYTES` does).

However, some will not like byte arrays thrashing, so they might bind `E` to an `okio.Buffer`, a `ByteCounter` for that would return its available bytes.

Thanks for asking the question.. it will help me write javadocs.

core/src/main/java/zipkin/reporter/AsyncReporter.java
```diff
@@ -0,0 +1,253 @@
...
+      message = format("%s due to %s(%s)", message, e.getClass().getSimpleName(),
+          e.getMessage() == null ? "" : e.getMessage());
+      logger.log(WARNING, message, e);
+    }
+  }
+
+  @AutoValue
+  static abstract class EncodingAsyncReporter<S, B> extends AsyncReporter<S> {
+
+    abstract Encoder<S, B> encoder();
```
adriancole: @felixbarny here's where I'd add javadoc explaining that this thing ends up controlling the element type of the bounded queue, using examples of `S` being `zipkin.Span` and `B` being `byte[]`.

core/src/main/java/zipkin/reporter/Sender.java
```diff
@@ -15,41 +15,45 @@
...
+ * <p>If performance is critical, {@link Encoding#THRIFT thrift encoding} is most efficient,
+ * sometimes an order of magnitude faster than json.
  *
- * @see SpanEncoder
- * @see ListEncoder
+ * @param <S> type of span to send. Usually an {@link Encoder#encode(Object) encoded buffer}.
  */
-// @FunctionalInterface
-public interface Sender extends Component {
+public interface Sender<S> extends Component {
```
adriancole: self-note: change this to B!

nicmunroe:
@adriancole LGTM so far. The simplicity of the example `WingtipsSpanReporter` you created tells me it's good. Proper separation of concerns with interfaces defining the contracts leads to easy integrations. :+1: 

> cc @nicmunroe don't pay too much attention to the details of the code, as I've another day of work cleaning it up. That said, lemme know if the approach and the api integration looks alright.

adriancole:
going to do a big refactor again. by removing the ability to create a non-encoding reporter, the whole codebase is simpler (ex makes things like what's a buffer or not extremely obvious).

after that, I'll focus on ensuring messages are possible to send. For example, we often have issues due to kafka `message.max.bytes` where producers inadvertently send more than possible to consume. This happens with scribe, too, just less frequently.

adriancole:
ok code is now simplified a lot. There's only one type of constraint for sending, which is the total bytes per message. This is much more appropriate for reporters as they don't care what is inside most of the time.

adriancole:
OK reasonably happy with the api at this point.

* sender decides how much it can accept (messageMaxBytes)
* reporter honors that by not dequeuing more than this at a time

Now's details time.. mostly backfilling tests

adriancole:
Added first tests (also added the ability to control worker count as that exists in htrace)

adriancole:
pulled sender-specific changes into a separate PR as they are relatively stable:  https://github.com/openzipkin/zipkin-reporter-java/pull/8

adriancole:
After benchmarking, I'm confident in the async approach. With default serialization overhead, `AsyncReporter.report(Span)` takes about 2microseconds on my laptop. I've tested that even with high contention (8 threads reporting), no spans are dropped and the reporting throughput is linear. 

I'm going to backfill more tests and plan to merge in the next day or two. fyi @nicmunroe @felixbarny

```
Benchmark                                                        Mode  Cnt      Score     Error   Units
AsyncReporterBenchmarks.high_contention                         thrpt   15   1550.877 ± 128.986  ops/ms
AsyncReporterBenchmarks.high_contention:high_contention_flush   thrpt   15    158.727 ±  13.241  ops/ms
AsyncReporterBenchmarks.high_contention:high_contention_report  thrpt   15   1392.150 ± 115.747  ops/ms
AsyncReporterBenchmarks.high_contention:messagesDropped         thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.high_contention:messagesSent            thrpt   15    158.745 ±  13.234  ops/ms
AsyncReporterBenchmarks.high_contention:spansDropped            thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.high_contention:spansReported           thrpt   15  11139.567 ± 924.662  ops/ms
AsyncReporterBenchmarks.mild_contention                         thrpt   15   1143.694 ±  21.488  ops/ms
AsyncReporterBenchmarks.mild_contention:messagesDropped         thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.mild_contention:messagesSent            thrpt   15    172.181 ±   4.248  ops/ms
AsyncReporterBenchmarks.mild_contention:mild_contention_flush   thrpt   15    172.179 ±   4.248  ops/ms
AsyncReporterBenchmarks.mild_contention:mild_contention_report  thrpt   15    971.515 ±  17.951  ops/ms
AsyncReporterBenchmarks.mild_contention:spansDropped            thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.mild_contention:spansReported           thrpt   15   1943.060 ±  35.909  ops/ms
AsyncReporterBenchmarks.no_contention                           thrpt   15    734.925 ±  25.125  ops/ms
AsyncReporterBenchmarks.no_contention:messagesDropped           thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.no_contention:messagesSent              thrpt   15    183.372 ±  15.215  ops/ms
AsyncReporterBenchmarks.no_contention:no_contention_flush       thrpt   15    183.372 ±  15.215  ops/ms
AsyncReporterBenchmarks.no_contention:no_contention_report      thrpt   15    551.553 ±  17.670  ops/ms
AsyncReporterBenchmarks.no_contention:spansDropped              thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.no_contention:spansReported             thrpt   15    551.557 ±  17.669  ops/ms
```

adriancole:
Measured that there's no notable backlog even at very high reporting speeds (10spans/us which no zipkin system can consume). This means we won't ever need more than one thread to invoke senders. However, we do need to be careful about senders who aren't asynchronous as they can block the sender loop. I made a note of this in javadoc and this is something we can followup with on a per-sender basis.

```
Benchmark                                                 Mode  Cnt      Score     Error   Units
AsyncReporterBenchmarks.high_contention                  thrpt   15   1252.066 ±  89.663  ops/ms
AsyncReporterBenchmarks.high_contention:messages         thrpt   15      2.795 ±   0.200  ops/ms
AsyncReporterBenchmarks.high_contention:messagesDropped  thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.high_contention:spanBacklog      thrpt   15      0.002 ±   0.009  ops/ms
AsyncReporterBenchmarks.high_contention:spans            thrpt   15  10017.368 ± 717.519  ops/ms
AsyncReporterBenchmarks.high_contention:spansDropped     thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.mild_contention                  thrpt   15    933.658 ±  62.117  ops/ms
AsyncReporterBenchmarks.mild_contention:messages         thrpt   15      0.521 ±   0.035  ops/ms
AsyncReporterBenchmarks.mild_contention:messagesDropped  thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.mild_contention:spanBacklog      thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.mild_contention:spans            thrpt   15   1867.322 ± 124.236  ops/ms
AsyncReporterBenchmarks.mild_contention:spansDropped     thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.no_contention                    thrpt   15    551.589 ±  41.728  ops/ms
AsyncReporterBenchmarks.no_contention:messages           thrpt   15      0.154 ±   0.012  ops/ms
AsyncReporterBenchmarks.no_contention:messagesDropped    thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.no_contention:spanBacklog        thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.no_contention:spans              thrpt   15    551.589 ±  41.728  ops/ms
AsyncReporterBenchmarks.no_contention:spansDropped       thrpt   15        ≈ 0            ops/ms
```

PR #8, opened 2016-08-25
Decouples Sender from byte[], adds message constraints and health check

adriancole:
Many libraries use byte buffers that aren't byte arrays. This change
allows flexibility for using other types like `ByteBuffer`.

This also allows senders to indicate their max message size, something
particularly helpful in Kafka.

Finally, this adds a health check (via `zipkin.Component`). This allows
you to fail fast, for example if Kafka metadata queries fail.


adriancole:
unleashed into v0.3.0

PR #10, opened 2016-08-29
Adds OkHttp sender

adriancole:
Until now, the only http transport was URLConnection, which has no
dependencies and JRE 6 runtime capability, but isn't asynchronous or
http/2 capable. For example, the URLConnection sender can easily block
the reporter thread and cause a backlog.

This adds an OkHttp sender with parameters for controlling concurrency.

### DIFF
--- a/okhttp3/pom.xml
+++ b/okhttp3/pom.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2016 The OpenZipkin Authors
+
+    Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+    in compliance with the License. You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software distributed under the License
+    is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+    or implied. See the License for the specific language governing permissions and limitations under
+    the License.
+
+-->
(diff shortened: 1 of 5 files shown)


adriancole:
I'm not done this, yet, as I need a fully integrated async reporter benchmark to put this stuff in perspective.

Also, the bench vs normal json isn't fair. I need to switch the normal one to not use reflection. If a fair comparison doesn't show okio-backed span-at-a-time encoder winning, I might yank it completely, and squash `B encode(S span)` to `byte[] encode(S span)`. We'll see...

adriancole:
@nicmunroe @kristofa @felixbarny fyi

benchmarks/src/main/java/zipkin/reporter/HttpSenderBenchmarks.java
```diff
@@ -0,0 +1,127 @@
...
+import zipkin.TestObjects;
+import zipkin.reporter.internal.AwaitableCallback;
+
+@Measurement(iterations = 5, time = 1)
+@Warmup(iterations = 10, time = 1)
+@Fork(3)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+@State(Scope.Group)
+public abstract class HttpSenderBenchmarks {
```
adriancole: @anuraaga still banging on this class, as there's value testing close to the code, but think "actual" benchmarks will need to end up jbender or grinder etc.

adriancole:
pulled benchmarks into a separate commit

adriancole:
for those interested benchmarks are in https://github.com/openzipkin/zipkin-reporter-java/pull/12

PR #11, opened 2016-08-31
Removes the B generic parameter from Encoder and AsyncReporter

adriancole:
Before, we had 3 generic type parameters:
* S - span model type
* B - buffer that ^^ encodes into, usually byte[]
* M - message type that a list of spans are encoded into, usually byte[]

The second parameter, B, was there to support more performant encoding,
for example, using pooled buffers. The theory was that since our json
encoder was slow, people might want to write their own.

As of zipkin 1.8, encoding speed of byte[] is exceptional, hundreds of
nanoseconds for a local span to a few microseconds for a complicated
json one. There is no longer a strong performance reason to encode spans
into something besides a byte[].

By removing the `B` generic parameter, we are left with `S` and `M`. The
message type is still valuable. For example, you can stream writes
instead of concatenating byte arrays, and of course Scribe requires us
to split spans into log-lines. `S` is needed due to other libraries
having different forms of the Span type, for example, thrift generated
classes.

Bottom-line is that we get a cognitive win by removing `B`, yet still
have flexibility in how to form messages.


adriancole:
Aside: with high contention, default encoders and a single message-bunding thread, we can encode more spans than most zipkin installations can collect. json throughput is about half thrift's in latest code (was 1/10th or so before).

```
Benchmark                                                         (encoding)   Mode  Cnt      Score     Error   Units
AsyncReporterBenchmarks.high_contention                                 JSON  thrpt   15    859.558 ±  44.926  ops/ms
AsyncReporterBenchmarks.high_contention:messages                        JSON  thrpt   15      5.428 ±   0.284  ops/ms
AsyncReporterBenchmarks.high_contention:messagesDropped                 JSON  thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.high_contention:spanBacklog                     JSON  thrpt   15      0.025 ±   0.072  ops/ms
AsyncReporterBenchmarks.high_contention:spans                           JSON  thrpt   15   6877.260 ± 359.321  ops/ms
AsyncReporterBenchmarks.high_contention:spansDropped                    JSON  thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.high_contention                               THRIFT  thrpt   15   1853.571 ± 118.346  ops/ms
AsyncReporterBenchmarks.high_contention:messages                      THRIFT  thrpt   15      5.694 ±   0.364  ops/ms
AsyncReporterBenchmarks.high_contention:messagesDropped               THRIFT  thrpt   15        ≈ 0            ops/ms
AsyncReporterBenchmarks.high_contention:spanBacklog                   THRIFT  thrpt   15      0.516 ±   0.926  ops/ms
AsyncReporterBenchmarks.high_contention:spans                         THRIFT  thrpt   15  14829.219 ± 946.680  ops/ms
AsyncReporterBenchmarks.high_contention:spansDropped                  THRIFT  thrpt   15      2.932 ±  12.141  ops/ms
```

adriancole:
@nicmunroe @felixbarny @shakuzen I think this is a positive simplification. lemme know if you feel otherwise!

nicmunroe:
I haven't gone over the changes here in detail, but the explanation comment sounds good to me! :+1:

adriancole:
thx for the review folks

PR #13, opened 2016-09-02
Removes the M generic parameter

adriancole:
After implementing the third reporter, I noticed that there's no value
exposing the type of message spans are bundled into. This changes the
api to eliminate the `Sender<M>` parameter, and fixes the message size
function to work with scribe which envelopes messages vs list concat.


README.md
```diff
@@ -46,7 +46,7 @@ Property | Description
 `messageMaxBytes` | Maximum bytes sendable per message including overhead. Default `Sender.messageMaxBytes`
 `messageTimeout` |  Maximum time to wait for messageMaxBytes to accumulate before sending. Default 1 second
 
-#### Don't block the flusher thread
```
adriancole: small adjustment which is unrelated.. it isn't bad to only use one thread and block during sends. It is only bad if you end up blocked to the point where the queue backs up. For example, brave and many others only use 1 sender thread and a blocking sender and they're fine.

adriancole:
@felixbarny fyi. after writing scribe support (on another branch) I realized the `M` type was over-engineering. so we now have only one parameter `S`, which we can't really do without! once this is in, I'll cut a release.

PR #17, opened 2016-09-17
Updates ReporterMetrics with stats used by Finagle and Jaeger

adriancole:
This adds `ReporterMetrics.updateSpanBacklog` and adds the cause for
`ReporterMetrics.incrementDroppedMessages`. These allow AsyncReporter
to produce metrics used by Finagle and Jaeger's tracers.

This is partially paid for by removing the metrics scope functionality.
Especially a client library shouldn't have a concern like this (even if
we do on the server).

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Here are the most important properties to understand when tuning.
 
 Property | Description
 --- | ---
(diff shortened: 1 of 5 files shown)


adriancole:
cc @felixbarny @nicmunroe

PR #20, opened 2016-09-29
WIP Amazon SQS reporter

llinder:
An Amazon SQS reporter that writes Spans as binary message attributes.

This is WIP port of an internal SQS Brave span collector.

### DIFF
--- a/awssqs/src/main/java/zipkin/reporter/awssqs/AwsBufferedSqsSender.java
+++ b/awssqs/src/main/java/zipkin/reporter/awssqs/AwsBufferedSqsSender.java
@@ -1,0 +1,203 @@
+/**
+ * Copyright 2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.reporter.awssqs;
+
+import com.amazonaws.ClientConfiguration;
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.AWSCredentialsProviderChain;
+import com.amazonaws.auth.AnonymousAWSCredentials;
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.auth.DefaultAWSCredentialsProviderChain;
+import com.amazonaws.auth.EnvironmentVariableCredentialsProvider;
+import com.amazonaws.auth.InstanceProfileCredentialsProvider;
+import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider;
+import com.amazonaws.auth.SystemPropertiesCredentialsProvider;
+import com.amazonaws.auth.profile.ProfileCredentialsProvider;
+import com.amazonaws.handlers.AsyncHandler;
+import com.amazonaws.internal.StaticCredentialsProvider;
+import com.amazonaws.services.sqs.AmazonSQSAsync;
+import com.amazonaws.services.sqs.AmazonSQSAsyncClient;
+import com.amazonaws.services.sqs.buffered.AmazonSQSBufferedAsyncClient;
+import com.amazonaws.services.sqs.model.GetQueueAttributesResult;
+import com.amazonaws.services.sqs.model.MessageAttributeValue;
+import com.amazonaws.services.sqs.model.SendMessageRequest;
+import com.amazonaws.services.sqs.model.SendMessageResult;
+import com.google.auto.value.AutoValue;
+import java.nio.ByteBuffer;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.apache.http.client.CredentialsProvider;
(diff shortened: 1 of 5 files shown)


awssqs/src/main/java/zipkin/reporter/awssqs/AwsBufferedSqsSender.java
```diff
@@ -0,0 +1,235 @@
...
+      return CheckResult.failed(e);
+    }
+  }
+
+  @Override protected AmazonSQSBufferedAsyncClient compute() {
+    // bound the executor so that we get consistent performance
+    ThreadPoolExecutor dispatchExecutor =
+        new ThreadPoolExecutor(0, maxRequests(), 60, TimeUnit.SECONDS,
+            new ArrayBlockingQueue<>(maxQueuedRequests()),
+            threadFactory("AwsBufferedSqsSender Dispatcher", false));
```
devinsba: This should probably be configurable from the builder

awssqs/src/main/java/zipkin/reporter/awssqs/AwsBufferedSqsSender.java
```diff
@@ -0,0 +1,235 @@
...
+  @Nullable abstract String awsServiceId();
+  abstract boolean compressionEnabled();
+  abstract int maxRequests();
+  abstract int maxQueuedRequests();
+
+  /** close is typically called from a different thread */
+  private AtomicBoolean closeCalled = new AtomicBoolean(false);
+
+
+  @Override public CheckResult check() {
```
devinsba: I understand the idea of this check but it requires more permissions than is required to send messages which will break when people try to lock things down to only the operations they intuitively think they will need.

I don't know a good way to check if you can write to a SQS queue though

awssqs/src/main/java/zipkin/reporter/awssqs/AwsBufferedSqsSender.java
```diff
@@ -0,0 +1,235 @@
...
+      credentials = credentialsProviderChain.getCredentials();
+      credentialsProvider =
+          new STSAssumeRoleSessionCredentialsProvider(credentials, roleArn, roleSessionName);
+    }
+
+    return credentialsProvider;
+  }
+
+  private static ThreadFactory threadFactory(final String name, final boolean daemon) {
+    return (Runnable r) -> {
```
devinsba: Unfortunately we cannot use Java 8 features in the reporters

awssqs/src/main/java/zipkin/reporter/awssqs/AwsBufferedSqsSender.java
llinder: Agreed.  I was at a loss for a better way to perform health checks.  Requiring extra permissions for this isn't the best.  I will either remove it or make the check optional through config.

awssqs/src/main/java/zipkin/reporter/awssqs/AwsBufferedSqsSender.java
llinder: Is your thought to make the actual ThreadPoolExecutor a configuration value or just the factory name?  What I have here was just modeled after the OkHttpSender.  Making it different is ok if thats preferable.

awssqs/src/main/java/zipkin/reporter/awssqs/AwsBufferedSqsSender.java
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  @Override public void close() {
+    if (!closeCalled.getAndSet(true)) {
+      AmazonSQSBufferedAsyncClient maybeNull = maybeNull();
+      if (maybeNull != null) maybeNull.shutdown();
+    }
+  }
+
+  private static AWSCredentialsProvider credentialsProvider(
```
llinder: Agreed.  I will make the change.

awssqs/src/main/java/zipkin/reporter/awssqs/AwsBufferedSqsSender.java
llinder: The zipkin-reporter-java project uses the Maven Retrolambda plugin.  https://github.com/openzipkin/zipkin-reporter-java/blob/master/pom.xml#L208  I've never used this but my understanding is that it converts lambda expressions to Java6 compatible byte code.

devinsba: ah, I've never heard of that. Interesting. Will look into it

awssqs/src/main/java/zipkin/reporter/awssqs/AwsBufferedSqsSender.java
devinsba: The `ThreadPoolExecutor`, if we provide a default one beyond the default one included in the aws-sdk then it should probably be configurable

awssqs/src/main/java/zipkin/reporter/awssqs/AwsBufferedSqsSender.java
```diff
@@ -0,0 +1,235 @@
...
+    if (closeCalled.get()) throw new IllegalStateException("closed");
+
+    checkNotNull(encodedSpans, "list of encoded spans must not be null");
+
+    for (byte[] span : encodedSpans) {
+      get().sendMessageAsync(new SendMessageRequest(queueUrl(), "z")
+              .addMessageAttributesEntry("z", new MessageAttributeValue()
+                  .withDataType("Binary." + encoding().name())
+                  .withBinaryValue(ByteBuffer.wrap(span))), new AsyncHandlerAdapter(callback));
+    }
```
devinsba: Could you paste an example of what this payload looks like on the receiving end or in the SQS console. I'm trying to decide what I need an SNS payload to look like so that the collector will be able to distinguish between the 2 payload types.

devinsba: Also you can send all the encoded spans as one message that matches the encoding using `BytesMessageEncoder.forEncoding(encoding()).encode(list)` to create a byte array that forms a list in the appropriate way

llinder: Thanks.  Updated using the suggested approach.

llinder:
Moved PR to new zipkin-aws repository: https://github.com/openzipkin/zipkin-aws/pull/3

adriancole:
Ps if you use retro lambda plugin in your new repo you can use lambda!  The
plug in rewrites byte code to java6

adriancole:
I wouldn't require a dependence on aws sdk if you can avoid it. This can
lead to class path peril later.

PR #23, opened 2016-10-05
Allow specifying JSON URLConnectionSender encoding

schlosna:
The `zipkin.reporter.urlconnection.AutoValue_URLConnectionSender.Builder#encoding(zipkin.reporter.Encoding)` method was previously not `public`, so one could not specify `JSON` as the encoding type. This PR makes that method accessible and updates tests to accommodate JSON encoding.

### DIFF
--- a/urlconnection/src/test/java/zipkin/reporter/urlconnection/URLConnectionSenderTest.java
+++ b/urlconnection/src/test/java/zipkin/reporter/urlconnection/URLConnectionSenderTest.java
@@ -14,14 +14,18 @@
 package zipkin.reporter.urlconnection;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.RecordedRequest;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 import zipkin.Span;
 import zipkin.TestObjects;
 import zipkin.junit.HttpFailure;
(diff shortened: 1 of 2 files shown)


urlconnection/src/test/java/zipkin/reporter/urlconnection/URLConnectionSenderTest.java
```diff
@@ -184,7 +233,19 @@ void thenCallbackCatchesTheThrowable() {
   /** Blocks until the callback completes to allow read-your-writes consistency during tests. */
   void send(List<Span> spans) {
     AwaitableCallback callback = new AwaitableCallback();
-    sender.sendSpans(spans.stream().map(Encoder.THRIFT::encode).collect(toList()), callback);
```
adriancole: I think it would be simpler to make Encoder a field in the test, and just change to map (encoder::encoding) right?

urlconnection/src/test/java/zipkin/reporter/urlconnection/URLConnectionSenderTest.java
```diff
@@ -161,6 +183,33 @@ public void illegalToSendWhenClosed() throws Exception {
     send(TestObjects.TRACE);
   }
 
+  @Test
+  public void jsonEncoding() throws Exception {
```
adriancole: If checking json encoding I wouldn't copy paste the assertions that don't have to do with json. Ex just check the sender.encoding and mediatype then send once and check they were received via stats (no need to check pre-send and second send as they don't have anything to do with json). This keeps the code more to the point and.. shorter!

schlosna: I had these as initial sanity checks, can remove them.

urlconnection/src/test/java/zipkin/reporter/urlconnection/URLConnectionSenderTest.java
schlosna: done

adriancole:
hoping this sorts out the flake. we'll see! https://github.com/openzipkin/zipkin-reporter-java/commit/8cf00bf4b8497c9febcfe08a8b287ea3cdf840e1

PR #24, opened 2016-10-13
When a LazyCloseable delegates to a Closeable, prefer default close()

adriancole:
`LazyCloseable.close()` is not final as there's no requirement that its
delegate implements Closeable (ex it could just have a `shutdown()`
method). Due to copy/pasta, we redundantly did the same thing in a
couple subtypes.

Thanks to @autoletics for pointing out this possibility.


adriancole:
bumped the general topic to its own issue

PR #25, opened 2016-10-14
Use daemon-thread to avoid blocking shutdown when close is not called

NegatioN:
It can be difficult to call close on AsyncReporter when the usage-context is unaware of when shutdown is about to happen. The current non-daemon thread implementation blocks shutdown when close isn't called.
We have batch-jobs run by cron that hangs due to the current implementation. 
If there is a reason for it not being a daemon-thread, we would like to propose to add the shutDownHook like below, as it should be safe to assume if the flusher thread is the last non-daemon thread left it can be stopped.

So this is a commit with both solutions included.
I'll revise the PR after feedback @adriancole

### DIFF
--- a/core/src/main/java/zipkin/reporter/AsyncReporter.java
+++ b/core/src/main/java/zipkin/reporter/AsyncReporter.java
@@ -147,7 +147,7 @@ public abstract class AsyncReporter<S> implements Reporter<S>, Flushable, Compon
       if (messageTimeoutNanos > 0) { // Start a thread that flushes the queue in a loop.
         final BufferNextMessage consumer =
             new BufferNextMessage(sender, messageMaxBytes, messageTimeoutNanos);
-        new Thread(() -> {
+        final Thread flushThread = new Thread(() -> {
           try {
             while (!result.closed.get()) {
               result.flush(consumer);
@@ -156,7 +156,9 @@ public abstract class AsyncReporter<S> implements Reporter<S>, Flushable, Compon
             for (byte[] next : consumer.drain()) result.pending.offer(next);
             result.close.countDown();
           }
-        }, "AsyncReporter(" + sender + ")").start();
+        }, "AsyncReporter(" + sender + ")");
+        flushThread.setDaemon(true);
+        flushThread.start();
       }
       return result;
     }


NegatioN:
@kristofa @michaelsembwever @jplock
Do any of you have time to look over this change?

core/src/main/java/zipkin/reporter/AsyncReporter.java
```diff
@@ -156,7 +156,9 @@ public Builder queuedMaxBytes(int queuedMaxBytes) {
             for (byte[] next : consumer.drain()) result.pending.offer(next);
             result.close.countDown();
           }
-        }, "AsyncReporter(" + sender + ")").start();
+        }, "AsyncReporter(" + sender + ")");
+        flushThread.setDaemon(true); //Let applications terminate even if thread is not properly closed.
```
michaelsembwever: there's no need for the trailing comment here – we know what daemon thread implies (or can just the apidoc on it)

core/src/main/java/zipkin/reporter/AsyncReporter.java
```diff
@@ -147,7 +147,7 @@ public Builder queuedMaxBytes(int queuedMaxBytes) {
       if (messageTimeoutNanos > 0) { // Start a thread that flushes the queue in a loop.
         final BufferNextMessage consumer =
             new BufferNextMessage(sender, messageMaxBytes, messageTimeoutNanos);
-        new Thread(() -> {
+        final Thread flushThread = new Thread(() -> {
```
michaelsembwever: i was going to say this project had to be Java6 compatible, but then discovered its use of retrolambda – how cool is that?!

michaelsembwever:
LGTM

NegatioN:
Rebased to one commit, and removed the superfluous comment. @michaelsembwever :)

michaelsembwever:
:+1:

michaelsembwever:
@adriancole :shipit:   (i don't have commits right on this repo)

kristofa:
👍

adriancole:
The rationale should actually be the reverse, IMHO. Especially where close is exposed, implicit threads should be non daemon unless proven otherwise. 

When we decide to make something a daemon thread due to the inconvenience of closing resources, it might lead to other problems (like state exceptions in logs or leaked network connections). When these later become a problem, the support issue is worse... ie paste of OOM error in gitter and hours spent trying to help them figure out why.

It is more typical to use daemon threads for cleanup tasks, or when you don't have a direct reference. For example, okhttp's connection pool has a shared static executor which makes it tough to reason with when considering how to close it, and the usage of the pool (which lingers around when not close) is well defined. In this case, it is pretty-much text book daemon thread.
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/ConnectionPool.java#L48

All this said, I do think folks here are aware of the pros and cons, and these are the same folks supporting issues like this. So, yeah I'm cool with this as long as our "official" contract remains Closeable!

Thanks for pitching the case and the code remedy.

adriancole:
ps sorry about write access, I forgot that we had to explicitly add teams to each repo "core" is now added.

michaelsembwever:
Agreed with @adriancole notion that proper design permits and uses resources that are programmatically closed.

PR #30, opened 2016-12-01
Fixes #28

### LINKED ISSUE #28: Test scope for io.zipkin.java:zipkin:test-jar
When I have a dependency on zipkin-sender-urlconnection it dependency manages in io.zipkin.java:zipkin:test-jar as compile scope.

As the dependency uses a different package type, it's not possible to exclude that dependency with Maven.

Should the dependency be in test scope, so as not to bring it in at compile time?

kenfinnigan:
Adds test scope to io.zipkin.java:zipkin:test-jar and adds it as a compile scope into /benchmarks

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -61,6 +61,14 @@
     </dependency>
 
     <dependency>
+      <groupId>io.zipkin.java</groupId>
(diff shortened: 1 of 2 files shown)


adriancole:
gracias!

PR #43, opened 2017-04-28
Limit toString() output for Sender implementations

### LINKED ISSUE #42: Improve name of AsyncReporter's flush thread
The flush thread created by AsyncReporter uses a thread name that includes the toString() output of the sender (see https://github.com/openzipkin/zipkin-reporter-java/blob/master/core/src/main/java/zipkin/reporter/AsyncReporter.java#L159). At least for the KafkaSender, this can result in very long thread names due to the inclusion of the Kafka producer configuration properties in KafkaSender's toString() output. In cases where the Kafka cluster being used requires TLS encryption or client authentication, inclusion of these properties in the thread name risks exposing sensitive configuration in log output.

Here is an example thread name I observed in logs using the 0.7.0 release and https://github.com/openzipkin/zipkin-reporter-java/blob/master/kafka10/src/main/java/zipkin/reporter/kafka10/KafkaSender.java as the sender. Some sensitive values have been replaced with placeholders.

```
AsyncReporter(KafkaSender{encoding=THRIFT, messageMaxBytes=1000000, encoder=THRIFT, topic=zipkin, properties={bootstrap.servers=192.168.0.1:9093,192.168.0.2:9093,192.168.0.3:9093, ssl.truststore.location=/not/a /real/path/trust.jks, acks=1, ssl.truststore.password=fakepassword, value.serializer=org.apache.kafka.common.serialization.ByteArraySerializer, security.protocol=SSL, key.serializer=org.apache.kafka.common.serialization.ByteArraySerializer}})
```

The length of the thread name makes reading or viewing logs awkward.

The inclusion of the value for ssl.truststore.password exposes information that should not be in logs. It is also not easy to excluded from log output, since thread name is generally useful and whether to include it in logs is often not configurable at the individual logger level.

Including properties is also redundant with the Kafka producer's logging. It will log it's configuration when the producer is created, replacing sensitive values with `[hidden]`.

Some possible ways to address this:
* Use the sender's class name in the thread name, instead of the sender's toString() output. This might reveal implementation details that could cause confusion, such as getting `AsyncReporter(AutoValue_KafkaSender)`, instead of `AsyncReporter(KafkaSender)`.
* Add a `String name();`, or similar, method to [Sender](https://github.com/openzipkin/zipkin-reporter-java/blob/master/core/src/main/java/zipkin/reporter/Sender.java) and use the value returned from that method in the thread name. This gives senders control over how to identify themselves, but requires existing Sender implementations to be updated to be compatible (unless Java 8 default methods are an option).
* Exclude the `properties` field from KafkaSender's toString output. This addresses the sensitive data concern and helps with the length concern. Results would look something like `AsyncReporter(KafkaSender{encoding=THRIFT, messageMaxBytes=1000000, encoder=THRIFT, topic=zipkin})`.

Are there other options? Is there a way to address this that doesn't require changing the implementation of AsyncReporter, and possibly other components?

dgrabows:
Changed implementations of zipkin.reporter.Sender to provide an explicit toString() implementation that only includes the classname of the sender. Previously, toString() was generated by AutoValue and included all fields.

The output of toString() on zipkin.reporter.Sender implementations appears in thread names created by zipkin.reporter.AsyncReporter. Since thread names are likely to be exposed in logs and other monitoring tools, these changes are meant to ensure the toString() output is a reasonable length for display in those contexts and does not contain sensitive information. One example of potentially sensitive information being include is the Kafka producer configuration properties previously output for KafkaSender. These properties can contain values such as keystore passwords.

The feedback on #42 was the "tighten up" the toString output. Are these changes too aggressive in limiting the output of toString? Leaving out all of the fields on Senders could make it hard to differentiate between AsyncReporter flush threads, if there are multiple running with different Senders of the same type, but that seems like an unlikely scenario. I can't think of a reason to have, e.g., two instances of KafkaSender running in the same service or application.

Fixes #42

### DIFF
--- a/local/src/test/java/zipkin/reporter/local/LocalSenderTest.java
+++ b/local/src/test/java/zipkin/reporter/local/LocalSenderTest.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016 The OpenZipkin Authors
+ * Copyright 2016-2017 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -53,6 +53,17 @@ public class LocalSenderTest {
     sender.close();
 
     send(TestObjects.TRACE);
+  }
+
+  /**
+   * The output of toString() on {@link zipkin.reporter.Sender} implementations appears in thread
+   * names created by {@link zipkin.reporter.AsyncReporter}. Since thread names are likely to be
+   * exposed in logs and other monitoring tools, care should be taken to ensure the toString()
+   * output is a reasonable length and does not contain sensitive information.
+   */
+  @Test
+  public void toStringContainsOnlySenderType() throws Exception {
+    assertThat(sender.toString()).isEqualTo("LocalSender");
   }
 
   /** Blocks until the callback completes to allow read-your-writes consistency during tests. */
--- a/local/src/test/java/zipkin/reporter/local/LocalSenderTest.java
+++ b/local/src/test/java/zipkin/reporter/local/LocalSenderTest.java
@@ -62,8 +62,9 @@ public class LocalSenderTest {
    * output is a reasonable length and does not contain sensitive information.
    */
   @Test
-  public void toStringContainsOnlySenderType() throws Exception {
-    assertThat(sender.toString()).isEqualTo("LocalSender");
+  public void toStringContainsOnlySenderTypeAndStorageType() throws Exception {
+    assertThat(sender.toString())
+        .isEqualTo("LocalSender(" + storage.getClass().getSimpleName() + ")");
   }
 
   /** Blocks until the callback completes to allow read-your-writes consistency during tests. */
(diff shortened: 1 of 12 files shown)


local/src/test/java/zipkin/reporter/local/LocalSenderTest.java
```diff
@@ -55,6 +55,17 @@ public void illegalToSendWhenClosed() throws Exception {
...
   }
 
+  /**
+   * The output of toString() on {@link zipkin.reporter.Sender} implementations appears in thread
+   * names created by {@link zipkin.reporter.AsyncReporter}. Since thread names are likely to be
+   * exposed in logs and other monitoring tools, care should be taken to ensure the toString()
+   * output is a reasonable length and does not contain sensitive information.
+   */
+  @Test
+  public void toStringContainsOnlySenderType() throws Exception {
```
adriancole: LocalSender(actual span store type) is probably more informative

adriancole:
ps thanks for taking a stab at this!

adriancole:
Thanks for helping, Dan. Will merge on green. Let me know when you need a release.

dgrabows:
Thanks for the quick feedback. Sometime in the month would be helpful, but isn't critical. I can always push a build of master or my fork to an internal repo in the meantime.

adriancole:
heh it is the last day of the month :D

dgrabows:
Sorry, I meant sometime in the _next_ month.

Issue #29, opened 2016-11-30
Is flush non-blocking?  How to ensure all spans are cleared?


dangets:
I am instrumenting a manually invoked jar and I am having troubles getting the traces to consistently show up.  I am explicitly calling `Reporter.flush()` before exiting.  I find that if I debug and give the app more time I am able to see the trace, but running normally I usually don't end up getting the entire trace.

Is there a blocking version of flush that I can use?

I _could_ use an `InMemoryReporterMetrics` and check the `queuedSpan` count, but this seems kind of hacky.

adriancole:
Correct. at the moment, when you close() the reporter, it blocks until the
last batch was sent, but the actual call to the sender is async. We could
adjust the code that when you call asyncReporter.close() block until the
sender operation succeeds or fails. Sound good?

adriancole:
well I suppose that or make a flag which when set any call to flush is
blocking (regardless of if the flush is manual or automatic)..

dangets:
Either way would work ... could it be a different builder option on the AsyncReporter like `flushSenderOnClose(boolean b)` or is that too specific to this situation? ... that may be what you were getting at with your first suggestion.

adriancole:
first suggestion was really to change the close() behavior as opposed to making a flag. close is already a blocking op

dangets:
Again, I'm good with either - do you have a preference?  I could try implementing if you'd like.

adriancole:
Why don't you give it a try!

here's some hints..

```java
    /** When true, calls to {@link #flush()} block until send operations complete. Default false */
    public Builder blockOnSend(boolean blockOnSend) {
      this.blockOnSend = blockOnSend;
      return this;
    }

...

        if (blockOnSend) {
          AwaitableCallback blockingCallback = new AwaitableCallback();
          sender.sendSpans(nextMessage, blockingCallback);
          blockingCallback.await(); // will propagate on exception
        } else {
          sender.sendSpans(nextMessage, failureCallback);
        }
```

Note that testing will be tricky. Also, whether or not we block, there still could be spans left in the queue which default behavior drops. For example, you could look at the close impl to try to clear it up to a specific timeout (or just assume all is ok)

dangets:
I added the code blocks you suggested, but am having some difficulty writing the tests.  In digging more into the code, I'm getting more confused why I had the original problem.

[...]

PR #46, opened 2017-05-10
Sends whatever we have on AsyncReporter.close

adriancole:
Before, we tried to fill up a message before returning from close. This
could end up blocking up to messageTimeout, especially if there aren't
many spans coming in.

Now, we send whatever we have. This also reduces the blocking time on
close to bounded at 1second vs whatever messageTimeout was. This is
because the latency is now related to how long it takes for a sender to
complete: we have no way of knowing this, so arbitrarily choose 1
second.

See #29

### DIFF
--- a/core/src/main/java/zipkin/reporter/AsyncReporter.java
+++ b/core/src/main/java/zipkin/reporter/AsyncReporter.java
@@ -69,6 +69,7 @@ public abstract class AsyncReporter<S> implements Reporter<S>, Flushable, Compon
     ReporterMetrics metrics = ReporterMetrics.NOOP_METRICS;
     int messageMaxBytes;
     long messageTimeoutNanos = TimeUnit.SECONDS.toNanos(1);
+    long closeTimeoutNanos = TimeUnit.SECONDS.toNanos(1);
     int queuedMaxSpans = 10000;
     int queuedMaxBytes = onePercentOfMemory();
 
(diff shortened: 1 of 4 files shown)


adriancole:
this was noticed by @sirtyro when setting messageTimeout to 30seconds. I verified this in a different unit test in brave, and this fix sorts that, too.

core/src/main/java/zipkin/reporter/AsyncReporter.java
```diff
@@ -244,13 +246,14 @@ void flush(BufferNextMessage bundler) {
 
     @Override
     public void close() {
-      closed.set(true);
+      if (!closed.compareAndSet(false, true)) return; // already closed
       try {
-        if (!close.await(messageTimeoutNanos, TimeUnit.NANOSECONDS)) {
-          logger.warning("Timed out waiting for close");
+        // wait a second for in-flight spans to send
+        if (!close.await(1, TimeUnit.SECONDS)) {
```
devinsba: Depending on the size of the buffer I'm not sure this would be enough time. Maybe it should be configurable aswell?

adriancole: how about `closeTimeout` defaulting to a second? or do you have a better default value in mind?

devinsba: possibly `closeTimeoutSeconds` so we can protect users from assuming its milliseconds

adriancole: ok in spring I'll suffix seconds (this has a parameter for timeunit)

adriancole: actually I guess in spring could go millis.. that's probably a better choice?

adriancole:
added closeTimeout here. thanks for the suggestion!

core/src/main/java/zipkin/reporter/AsyncReporter.java
sirtyro: Keep it consistent with https://github.com/openzipkin/brave/pull/409/files#r115881300

aka millis

PR #56, opened 2017-08-04
Bumps versions notably zipkin

adriancole:


urlconnection/src/main/java/zipkin/reporter/urlconnection/URLConnectionSender.java
```diff
@@ -162,24 +162,42 @@ void send(byte[] body, String mediaType) throws IOException {
...
+  static void skipAllContent(HttpURLConnection connection) throws IOException {
+    InputStream in = connection.getInputStream();
+    IOException thrown = skipAndSuppress(in);
+    if (thrown == null) return;
+    InputStream err = connection.getErrorStream();
+    if (err != null) skipAndSuppress(err); // null is possible, if the connection was dropped
+    throw thrown;
+  }
+
+  static IOException skipAndSuppress(InputStream in) {
```
adriancole: had to do this as errorprone is more strict about classes created by retrolambda

PR #62, opened 2017-09-10
Fixes #61 : add Util.class to shaded jar.

### LINKED ISSUE #61: Broken shaded internal in kafka sender
Bundled version of LazyCloseable in kafka-sender 1.0.2 is broken : 
```
...
   public void close() throws IOException {
      T maybeNull = this.maybeNull();
      if (maybeNull != null) {
         Util.checkArgument(maybeNull instanceof Closeable, "Override close() to close " + maybeNull, new Object[0]);
         ((Closeable)maybeNull).close();
      }
   }
...

```
In the close method, the Util class is not imported neither shaded in the jar. This leads to a ClassNotFoundException when calling this method.

hypnoce:


adriancole:
thx will merge this, then revert once zipkin 1.31.3 sorts the related problem

hypnoce:
Deal

PR #66, opened 2017-09-20
Updates to Zipkin v2, dropping Scribe, Kafka 0.8 and Local senders

adriancole:
This adds a v2 native sender which results in a lower dependency jar,
and cancelable senders. Along the way, this drops support for transports
including local, kafka08 and scribe. We also no longer support thrift
encoding. v1 code rarely changed, so old senders can be used until folks
can upgrade.


adriancole:
cc @openzipkin/core @llinder @devinsba @dgrabows @schlosna @cschneider this makes it about 150KiB total in dependencies for tracers that don't need to use v1 types

adriancole:
this uses maven group id `io.zipkin.reporter2` (as we need to be under io.zipkin)
the package name matches core zipkin, as `zipkin2.reporter`

adriancole:
will consider re-adding dropped transports if we get requests to!

PR #68, opened 2017-09-23
Fixes #67 : Make sure kafka producer is called when KafkaSender is cl…

### LINKED ISSUE #67: [Bug] Kafka producer is not closed when kafka11 sender is closed
Hi,

I realised that the kafka producer is no more a LazyCloseable. Therefore, the close method do not call the kafka producer close method and leaks into the Java MXBean registry.

I'll submit a test+fix soon.

Thanks

hypnoce:
…osed.

### DIFF
--- a/kafka11/src/main/java/zipkin2/reporter/kafka11/KafkaSender.java
+++ b/kafka11/src/main/java/zipkin2/reporter/kafka11/KafkaSender.java
@@ -15,6 +15,7 @@ package zipkin2.reporter.kafka11;
 
 import com.google.auto.value.AutoValue;
 import com.google.auto.value.extension.memoized.Memoized;
+import java.io.Closeable;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


kafka11/src/main/java/zipkin2/reporter/kafka11/KafkaSender.java
```diff
@@ -169,6 +170,10 @@ public final KafkaSender build() {
 
   @Override public void close() throws IOException {
     if (closeCalled) return;
+    final Closeable closeable = get();
```
adriancole: I think it is impossible for this to return null

adriancole:
doublechecked and I think this is the only sender with this bug. will address the small nit on the way in. Thanks!

adriancole:
releasing immediately

hypnoce:
Sorry I was sleeping :) Thanks ! (btw there is the Closeable import that you could remove :p)

PR #70, opened 2017-10-07
Adds amqp-client for RabbitMQ backend

adriancole:
This adds amqp-client for the recently added RabbitMQ support by @shakuzen

the library is named after the dependency

### DIFF
--- a/amqp-client/src/main/java/zipkin2/reporter/amqp/RabbitMQSender.java
+++ b/amqp-client/src/main/java/zipkin2/reporter/amqp/RabbitMQSender.java
@@ -23,7 +23,6 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.TimeoutException;
-import java.util.logging.Logger;
 import zipkin2.Call;
 import zipkin2.Callback;
 import zipkin2.CheckResult;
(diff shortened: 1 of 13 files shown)


adriancole:
OK should be better now.. added the travis service so that tests can run (also added a benchmark)

amqp-client/src/main/java/zipkin2/reporter/amqp/RabbitMQSender.java
```diff
@@ -0,0 +1,246 @@
...
+      this.message = message;
+    }
+
+    @Override protected Void doExecute() throws IOException {
+      publish();
+      return null;
+    }
+
+    void publish() throws IOException {
+      Channel channel = get().createChannel();
```
shakuzen: I think this is the simplest way to make this sender thread safe, but it does add two additional network roundtrips for every send (one to create the channel, one to close the channel). For better publishing performance, each thread should reuse a channel or a channel cache could be utilized. I doubt whether this would make a significant difference in the "normal" case where we expect sending to only happen every second, but under more pressure, it might matter.

adriancole:
Excellent feeback! I think I got it all, and the code is better now thanks
to you!

PR #73, opened 2017-10-14
Reverts change to use JSR 305 dependency, due to problems w/OSGi, Java 9

adriancole:
Before, we optimized to drop our source-retention Nullable in favor of
JSR 305. In doing so, we got null analysis from Intellij. However, in
doing so, we entered a swamp of dependency conflicts, which started with
OSGi (hence switch to service mix bundle) and now with Java 9 (conflict
with apps using jax-ws.

Since we are a core dependency, and particularly not a user-api, this
change caused more problems than it fixed. Hence, this reverts to the
safer model of using internal source-retention nullable (which allows
things like AutoValue to work, but don't introduce any chance of runtime
class conflicts).

See https://github.com/openzipkin/brave/issues/451


adriancole:
PS we never added JSR 305 to the 1.x line, so no problems there

PR #83, opened 2017-12-11
Adds spring beans XML config for legacy applications

adriancole:
This adds the ability to configure all properties for senders,
notably Kafka topic.

Fixes https://github.com/openzipkin/brave-webmvc-example/issues/19


adriancole:
This codebase can be used by any library, not just Brave. However, Brave will continue this effort, re-jigging its XML config support to Spring 2.x (from 3.x), using this so you can configure details like the kafka topic name https://github.com/openzipkin/brave/issues/549

Issue #81, opened 2017-12-07
Add a disruptor based async reporter


codefromthecrypt:
We can calculate the size of a span without serializing it. We should be able to make an async reporter that defers serialization, but still allows us to drop spans that are too big to send. It should be fun work, if someone's up to it.

We could also consider making this disruptor based, or otherwise with a fancier ringbuffer opt-in given a dep like https://github.com/real-logic/agrona.

wu-sheng:
In SkyWalking, we put the Spans(actually Segment) into a ringbuffer, then consume it, finally send it. So I agree with you, defer serialization is a good idea. 

For the ringbuffer, [LMAX-Exchange/disruptor](https://github.com/LMAX-Exchange/disruptor) is really a good project. For SkyWalking, we use a array based queue, implement by ourself.

adriancole:
Deferring serialization does help, but there's likely more to be saved by changing how the buffer works. Since our encoders are pretty good, we only save a few microseconds by deferring encoding:

There are brave's end-to-end benchmarks:

```
Benchmark                                           Mode  Cnt    Score    Error  Units
EndToEndBenchmarks.server_get                       avgt   15  133.966 ±  1.645  us/op
EndToEndBenchmarks.unsampledServer_get              avgt   15  159.102 ±  3.193  us/op

Marshaling up front:
EndToEndBenchmarks.traced128Server_get              avgt   15  201.900 ±  9.384  us/op

Deferred Marshaling:
EndToEndBenchmarks.traced128Server_get              avgt   15  198.209 ±  4.196  us/op
```

adriancole:
The above example records 2 spans, so there's overhead in low tens of microseconds per span.

There's no point in refactoring the existing async reporter, as the above deferred marshaling change was quite a lot of code. It isn't worth it for a small change. 

The async reporter needs to be completely rewritten to get above where we are at. It might be smarter to rewrite it at Brave's level as doing so could eliminate some lock contention and/or GC thrash around span updates.

adriancole:
whoops dirty workspace. Ignore my previous comments.

adriancole:
ignoring my other comment.. deferring is certainly worth doing, independent of this issue, which I've retitled to making things disruptor based https://github.com/openzipkin/zipkin-reporter-java/pull/84

adriancole:
[...]

PR #84, opened 2017-12-12
Reduces overhead by deferring codec to async thread

adriancole:
We calculate a span size in bytes before offering it to the async queue
to ensure we don't add a poison message (a span too big to encode).
However, we don't have to encode it on the application calling thread.

This operation reduces overhead by about 18us in benchmarking.

End-to-end benchmarks from Brave, running on jdk9.01
```
Benchmark                               Mode  Cnt    Score   Error  Units
EndToEndBenchmarks.server_get           avgt   15  133.535 ± 5.378  us/op
EndToEndBenchmarks.unsampledServer_get  avgt   15  145.541 ± 2.757  us/op

Before:
Benchmark                               Mode  Cnt    Score   Error  Units
EndToEndBenchmarks.traced128Server_get  avgt   15  207.332 ± 3.377  us/op

After:
Benchmark                               Mode  Cnt    Score   Error  Units
EndToEndBenchmarks.traced128Server_get  avgt   15  189.703 ± 3.342  us/op
```

See #81

### DIFF
--- a/amqp-client/src/main/java/zipkin2/reporter/amqp/RabbitMQSender.java
+++ b/amqp-client/src/main/java/zipkin2/reporter/amqp/RabbitMQSender.java
@@ -158,6 +158,10 @@ public abstract class RabbitMQSender extends Sender {
     return encoding().listSizeInBytes(encodedSpans);
   }
 
+  @Override public int messageSizeInBytes(int encodedSizeInBytes) {
+    return encoding().listSizeInBytes(encodedSizeInBytes);
+  }
+
   /** This sends all of the spans as a single message. */
   @Override public Call<Void> sendSpans(List<byte[]> encodedSpans) {
     if (closeCalled) throw new IllegalStateException("closed");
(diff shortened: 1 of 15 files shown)


adriancole:
will rebase over https://github.com/openzipkin/zipkin/pull/1838 once released

adriancole:
note: disruptor based support will be a separate issue https://github.com/openzipkin/zipkin-reporter-java/issues/81

adriancole:
@tramchamploo this had a huge impact in real-life.. might want to re-check some work you were doing early this year https://github.com/spring-cloud/spring-cloud-sleuth/pull/808

adriancole:
ps @tramchamploo I think was the first to call attention to the problem in the lock. I dismissed the concern instead of looking deeply at it. My mistake.

tramchamploo:
Is there a case for this impact? I think I'm experiencing this impact, but not sure if it is for this reason.

adriancole:
I used end to end benchmarks (in instrumentation/benchmarks) and also using
wrk to test this. Whether it helps with your app.. will have to test and
let us know. I hope it helps.

PR #86, opened 2017-12-12
Hardens deployment process

adriancole:


pom.xml
```diff
@@ -486,7 +496,7 @@
 
           <plugin>
             <artifactId>maven-javadoc-plugin</artifactId>
-            <version>2.10.4</version>
+            <version>3.0.0</version>
```
adriancole: needed for JDK 9!

PR #98, opened 2018-03-03
Adds zipkin-reporter-bom and uses it in spring beans and benchmarks

adriancole:
This adds a BOM which allows easy version alignment. As noted in the
README, you can use zipkin.version and zipkin-reporter.version to
change deps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,3 +104,50 @@ You can switch to v1 encoding like so:
 reporter = AsyncReporter.builder(URLConnectionSender.create("http://localhost:9411/api/v1/spans"))
                         .build(SpanBytesEncoder.JSON_V1);
 ```
+
+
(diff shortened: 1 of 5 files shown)


adriancole:
cc @jorgheymans @marcingrzejszczak @devinsba @spencergibb this is similar to what was recently merged in brave except it doesn't have a duplicate pom just to set license info and only for the bom (I will also remove this from brave). It also adds the variables used to override zipkin version easily

adriancole:
This would have been nice to be in brave 4.17 but 4.17.1 is fine. I won't announce 4.17.0 as it was an emergency release that went out mid-way through this BOM effort. I'll announce it after we get this in and brave using it.

adriancole:
tested with brave

PR #105, opened 2018-04-27
Enables proto3 encoding

adriancole:
By default, senders use json v2 encoding, which is easy to understand, but 
twice as large as binary encoding for normal spans. If you are running a
recent (2.8+) version of Zipkin server, consider [Protocol Buffers](https://developers.google.com/protocol-buffers/docs/proto3)
when you want to optimize for least size.

You can switch to proto3 encoding like so:

```java
sender = KafkaSender.newBuilder()
                    .encoding(Encoding.PROTO3)
                    .bootstrapServers("192.168.99.100:9092")
                    .build()
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,10 +98,26 @@ class CustomReporter implements Flushable {
   }
 ```
 
+
+### Protocol Buffers Encoding
+By default, senders use json v2 encoding, which is easy to understand, but
+twice as large as binary encoding for normal spans. If you are running a
+recent (2.8+) version of Zipkin server, consider [Protocol Buffers](https://developers.google.com/protocol-buffers/docs/proto3)
+when you want to optimize for least size.
+
+You can switch to proto3 encoding like so:
+
+```java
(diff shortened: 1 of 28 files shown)


adriancole:
3 hits to the brave-webmvc-example app
JSON_V2 "counter.zipkin_collector.bytes.http":2885.0
PROTO3 "counter.zipkin_collector.bytes.http":1251.0

adriancole:
3 hits to the brave-webmvc-example app (against kafka 1.1)
JSON_V2 "counter.zipkin_collector.bytes.kafka":2883.0
PROTO3 "counter.zipkin_collector.bytes.kafka":1253.0

adriancole:
3 hits to the brave-webmvc-example app (newer version of brave to kafka 0.8)
JSON_V2 "counter.zipkin_collector.bytes.kafka":3129.0
PROTO3 "counter.zipkin_collector.bytes.kafka":1497.0

PR #112, opened 2018-05-31
Switches java license format to code comment style

adriancole:
Otherwise latest code formatter adds javadoc tags to the license!


pom.xml
```diff
@@ -343,6 +343,10 @@
         <version>${license-maven-plugin.version}</version>
         <configuration>
           <header>${main.basedir}/src/etc/header.txt</header>
+          <mapping>
```
adriancole: here is the actual change

PR #114, opened 2018-06-13
Adds thrift encoding

adriancole:
This adds backported thrift encoding needed for Scribe and any kafka or http senders to old collectors.

### DIFF
--- a/core/src/main/java/zipkin2/reporter/BytesMessageEncoder.java
+++ b/core/src/main/java/zipkin2/reporter/BytesMessageEncoder.java
@@ -14,6 +14,7 @@
 package zipkin2.reporter;
 
 import java.util.List;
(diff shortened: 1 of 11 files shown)


core/src/main/java/zipkin2/reporter/BytesMessageEncoder.java
```diff
@@ -14,6 +14,7 @@
 package zipkin2.reporter;
 
 import java.util.List;
+import org.omg.PortableServer.THREAD_POLICY_ID;
```
shakuzen: 😱 OMG
I mean, is this import a mistake? 😝

adriancole: good catch. unused by needed to be removed! 11b612d5b54d6fd57f52b2b94e622a77768f3cbf

PR #121, opened 2018-08-21
Add AdminClient for health-checks

jeqo:
Following up https://github.com/openzipkin/zipkin/issues/2098, health checks has been added for KafkaSender.

### DIFF
--- a/kafka11/src/main/java/zipkin2/reporter/kafka11/KafkaSender.java
+++ b/kafka11/src/main/java/zipkin2/reporter/kafka11/KafkaSender.java
@@ -17,10 +17,14 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerConfig;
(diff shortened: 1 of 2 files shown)


kafka11/src/main/java/zipkin2/reporter/kafka11/KafkaSender.java
```diff
@@ -208,6 +214,18 @@ public Builder toBuilder() {
...
+
+  AdminClient getAdminClient() {
+    if (adminClient == null) {
+      synchronized (this) {
+        if (adminClient == null) {
+          adminClient = AdminClient.create(properties);
+        }
+      }
+    }
+    return adminClient;
```
adriancole: do we need to close this on close?

kafka11/src/main/java/zipkin2/reporter/kafka11/KafkaSender.java
```diff
@@ -230,6 +230,7 @@ AdminClient getAdminClient() {
     if (closeCalled) return;
     KafkaProducer<byte[], byte[]>  producer = this.producer;
     if (producer != null) producer.close();
+    if (adminClient != null) adminClient.close();
```
adriancole: since this is a volatile ref, the safe way is to make a local variable like the producer. then you don't have a risk of a NPE race!

adriancole:
thanks again!

PR #126, opened 2018-11-28
Bug: Fix the criterion for judging whether the span buffer is full

kapsterio:
It is nearly impossible to meet the the ```bufferFull``` criterion of BufferNextMessage, because the total span bytes size in buffer is almost impossible to be exactly equal to the ```maxBytes``` threshold. 

In our use case, we use KafkaSender with default value for ```maxBytes```(1000000 bytes),  when the span generation speed is over 1MiB per second, reporter began to drop spans, and the ```Messages Sended Per Second``` metric stabilized at 1/s due to buffer timeout.

The solution for this issue is quite straightforward, just set ```bufferFull``` to 'true' when the total span bytes size greater than or equal to ```maxBytes```.

### DIFF
--- a/core/src/main/java/zipkin2/reporter/BufferNextMessage.java
+++ b/core/src/main/java/zipkin2/reporter/BufferNextMessage.java
@@ -134,7 +134,10 @@ abstract class BufferNextMessage<S> implements SpanWithSizeConsumer<S> {
     int y = maxBytes;
     int includingNextVsMaxBytes = (x < y) ? -1 : ((x == y) ? 0 : 1); // Integer.compare, but JRE 6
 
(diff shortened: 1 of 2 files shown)


adriancole:
So, to play back:

Right now, the buffer waits until it is exactly full or a timeout occurs.
As messages are ordered, it is impossible to have a behavior besides
timeout because when a span tips over capacity, there's no way to get one
that wouldn't. This effectively means you have to wait until timeout. The
change here will flush on the first span will make it over capacity.

I'm surprised this bug stayed so long! Thanks for the help. Can you please
add a unit test so this stays fixed?

kapsterio:
Exactly as you describe, thanks for reviewing. 
I have already add test code for this case, check out when you are available.

adriancole:
great find. will send a patch out shortly

PR #128, opened 2019-02-21
Micrometer implementation of ReporterMetrics

### LINKED ISSUE #103: Feature request: Micrometer publisher for Reporter Metrics
It would be great to get a `ReporterMetrics` implementation that writes into Micrometer as the new metrics library in Spring Boot 2. This would also enable an autoconfiguration in Sleuth

If I get rule of 3 I'll clean up my in progress code as a PR

shakuzen:
Adds a Micrometer-based implementation of the `ReporterMetrics` interface. This will help users easily export reporter metrics to a backend where they can query and alert to find reporting issues.

Closes #103

---

Most of the credit goes to @devinsba for the bulk of the implementation.

### DIFF
--- a/metrics-micrometer/README.md
+++ b/metrics-micrometer/README.md
@@ -1,0 +1,23 @@
+Zipkin Reporter Metrics by Micrometer
+=====
+
+This module provides an implementation of Zipkin Reporter's `ReporterMetrics` for [Micrometer](https://micrometer.io).
+
+Usage
+---
+
+Pass an instance of `MicrometerReporterMetrics` to the `AsyncReporter` builder like below.
+
+```java
+reporter = AsyncReporter.builder(sender).metrics(MicrometerReporterMetrics.create(meterRegistry)).build();
+```
+
+### Extra tags
+
+You may optionally configure Micrometer tags to add to all reporter metrics created using the builder.
+
+For example, if an equivalent common tag is not already configured, you will want the Zipkin service name to be tagged so reporter metrics can be differentiated by service.
+
+```java
+MicrometerReporterMetrics reporterMetrics = MicrometerReporterMetrics.builder(meterRegistry).extraTags(reporterTags).build();
+```
--- a/metrics-micrometer/src/main/java/zipkin2/reporter/metrics/micrometer/MicrometerReporterMetrics.java
+++ b/metrics-micrometer/src/main/java/zipkin2/reporter/metrics/micrometer/MicrometerReporterMetrics.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright 2016-2019 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.reporter.metrics.micrometer;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import zipkin2.reporter.ReporterMetrics;
+
+import java.util.Arrays;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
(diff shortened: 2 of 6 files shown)


shakuzen:
This is ready for review if anyone has some spare cycles. (not sure if a notification is sent when I switched the pull request from "draft" to "ready to review")

metrics-micrometer/src/main/java/zipkin2/reporter/metrics/micrometer/MicrometerReporterMetrics.java
```diff
@@ -0,0 +1,123 @@
...
+  final AtomicInteger queuedBytes;
+
+  /**
+   * Uses the given {@link MeterRegistry} to make all the metrics.
+   * If any {@link Tag extraTags} are passed, they will be passed to all metrics this creates.
+   *
+   * @param meterRegistry all metrics will be registered to this registry
+   * @param extraTags     additional tags to attach to all metrics
+   */
+  public MicrometerReporterMetrics(MeterRegistry meterRegistry, Tag... extraTags) {
```
adriancole: not a fan of public constructors, as at least in spring it leads to tons of them. I'd use the factory/builder pattern like the other things. For example .create(meterRegistry) or .newBuilder(meterRegistry).extraTags(extraTags).build()

I'd be inclined to defer the extraTags feature unless there's a use case folks currently have for it.

shakuzen: > not a fan of public constructors, as at least in spring it leads to tons of them. I'd use the factory/builder pattern like the other things.

I'll update to switch to a Builder.

> I'd be inclined to defer the extraTags feature unless there's a use case folks currently have for it.

I added this to allow for https://github.com/openzipkin/zipkin-reporter-java/issues/103#issuecomment-440135719 (copied below) as well as any arbitrary tags someone might want to put on Zipkin reporter metrics.

> key enabler here would be a tag supplied by whatever the library's config is corresponding with Span.localEndpoint.serviceName

I suspect many will configure tags to distinguish applications/instances as common tags (applied to all metrics; not just Zipkin reporter metrics), which then wouldn't require anything additional here. If for some reason the service name for Zipkin is different from one such common tag, I thought it would be good to allow a way to define that. If we wanted to pick a tag name, we could offer an option in the builder specifically for this, but as I said, I suspect many will have it covered via a common tag already.

adriancole: sgtm

shakuzen: Put another way about extra tags: if the Zipkin service name happens to be different than the tag value used (if set at all) for the application/instance name, there should be a way to configure the service name tag for Zipkin reporter metrics without adding the (unrelated) tag to all other metrics in an application.

shakuzen:
Bob the Builder'd things.

metrics-micrometer/src/main/java/zipkin2/reporter/metrics/micrometer/MicrometerReporterMetrics.java
```diff
@@ -0,0 +1,154 @@
...
+    queuedSpans.set(i);
+  }
+
+  @Override
+  public void updateQueuedBytes(int i) {
+    queuedBytes.set(i);
+  }
+
+  public static final class Builder {
+    private MeterRegistry meterRegistry;
```
adriancole: final

metrics-micrometer/src/main/java/zipkin2/reporter/metrics/micrometer/MicrometerReporterMetrics.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Implementation of {@link ReporterMetrics} with Micrometer.
+ */
+public class MicrometerReporterMetrics implements ReporterMetrics {
+
+  private static final String PREFIX = "zipkin.reporter.";
+
+  private MeterRegistry meterRegistry;
```
adriancole: final this and below

metrics-micrometer/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+Pass an instance of `MicrometerReporterMetrics` to the `AsyncReporter` builder like below.
+
+```java
+reporter = AsyncReporter.builder(sender).metrics(MicrometerReporterMetrics.create(meterRegistry)).build();
+```
+
+### Extra tags
+
+You may optionally configure Micrometer tags to add to all reporter metrics created using the builder.
```
shakuzen: It would probably be good to mention here the one use case we're aware of for extra tags: Zipkin service name as a tag if not already present.

shakuzen: Added. Please check.

adriancole: sounds reasonable

PR #132, opened 2019-03-07
Fixes Kafka config to honor messageMaxBytes

adriancole:
Before this change, we'd actually overrun kafka's max message size due
to batching. This forces one message per batch and aligns config
properties. Tested against a real kafka server and
`RecordTooLargeException` went away.

### DIFF
--- a/kafka08/src/main/java/zipkin2/reporter/kafka08/KafkaSender.java
+++ b/kafka08/src/main/java/zipkin2/reporter/kafka08/KafkaSender.java
@@ -50,7 +50,7 @@ public final class KafkaSender extends Sender {
     properties.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, ByteArraySerializer.class.getName());
     properties.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG,
         ByteArraySerializer.class.getName());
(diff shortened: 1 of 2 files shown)


adriancole:
cc @jeqo @ImFlog

ImFlog:
Why do we need to batch if it's for one message ? Can't we just put 0 to disable batching ? (I haven't tested it).

adriancole:
if zero disables yeah that would work

adriancole:
@ImFlog was right. zero disables batching https://kafka.apache.org/documentation/

jeqo:
I was checking the KafkaProducer implementation and `RecordTooLarge` is evaluated per message: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L1015

I think disabling batching is not needed to enforce `max.request.size`, it could actually have good effects when many spans are produced in a period <= `linger.ms` and we reduce the round trips to Kafka. WDYT?

adriancole:
I will revert the batch size thing and then try again, but I was pretty sure I tried this and it still blew up

adriancole:
@jeqo you were right https://github.com/openzipkin/zipkin-reporter-java/pull/134

PR #134, opened 2019-03-08
Updates to latest zipkin and removes redundant batch setting

adriancole:
@jeqo noticed the prior change included a distracting batch setting that
wasn't necessary. The key fix there was correctly applying the max size,
not the batch size.


jeqo:
@adriancole awesome :) I was pushing hard my docker instances to generate force `RecordTooLarge` exception but I was only able to do it by applying `max.request.size`. 
I will work a PR with some test cases to validate this and document this behavior.

PR #135, opened 2019-03-08
Disabling kafka batching

jeqo:
Changes:
+ Disabling Kafka batching as already handled by Sender.

Nice to have:
+ Pass overrides as `Properties` instead of `Map` given that most Kafka configs are using this API.

### DIFF
--- a/kafka11/src/main/java/zipkin2/reporter/kafka11/KafkaSender.java
+++ b/kafka11/src/main/java/zipkin2/reporter/kafka11/KafkaSender.java
@@ -54,6 +54,9 @@ public final class KafkaSender extends Sender {
     properties.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, ByteArraySerializer.class.getName());
     properties.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG,
         ByteArraySerializer.class.getName());
+    // disabling batching as duplicates effort covered by sender buffering.
+    properties.put(ProducerConfig.BATCH_SIZE_CONFIG, 0);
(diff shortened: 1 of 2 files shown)


jeqo:
CC @adriancole @ImFlog 
Based on #132 and discussion about Kafka producer batching vs Zipkin buffering, disabling batching on Kafka producer make sense to avoid duplicating efforts.

kafka11/src/main/java/zipkin2/reporter/kafka11/KafkaSender.java
```diff
@@ -127,6 +134,29 @@ public final Builder overrides(Map<String, ?> overrides) {
...
+     * <p>For example: Reduce the timeout blocking from one minute to 5 seconds.
+     * <pre>{@code
+     * Properties overrides = new Properties();
+     * overrides.put(ProducerConfig.MAX_BLOCK_MS_CONFIG, 5000);
+     * builder.overrides(overrides);
+     * }</pre>
+     *
+     * @see ProducerConfig
+     */
+    public final Builder overrides(Properties overrides) {
```
adriancole: weird.. i thought properties would have matched on signature map string ?

jeqo: Yeap, it works with `Map<?,?>` as Properties implemets `Hashtable<Object, Object>` but that's a bit weirder IMHO. KafkaProducer has a `propsToMap`: https://github.com/apache/kafka/blob/201da0542726472d954080d54bc585b111aaf86f/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L1196

adriancole:
gotcha thanks!

PR #141, opened 2019-04-29
Does all the things to change to ASF build

adriancole:


adriancole:
thx for the looks folks

PR #151, opened 2019-06-25
Updates to latest versions and swaps undertow for armeria

adriancole:
We had a security vulnerability note about undertow. Since the community
is familiar with Armeria, this makes it a good time to swap. This change
only affects benchmarks.

cc @anuraaga @trustin @minwoox just for fun

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -92,9 +92,9 @@
     </dependency>
 
     <dependency>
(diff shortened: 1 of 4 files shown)


adriancole:
ran both benchmarks that used an http server (urlconnection and okhttp). looks correct as okhttp is slightly more throughput.

PR #156, opened 2019-08-07
Lightens the AwaitableCallback a bit

adriancole:
@Logic-32 and @anuraaga schooled me that when using Java concurrency
utilities in `java.util.concurrent`, you also inherit the thread
visibility properties ensured with synchronized and volatile symbols.

https://docs.oracle.com/javase/6/docs/api/java/util/concurrent/package-summary.html#MemoryVisibility

Given this, a volatile marker on a field whose state is dependent on
`CountDownLatch` should be redundant. This removes that volatile
qualifier accordingly.

cc other gear-heads @raphw @nicmunroe and @felixbarny for fun and good
measure!

see https://github.com/openzipkin/zipkin/pull/2735#discussion_r310878574

### DIFF
--- a/core/src/main/java/zipkin2/reporter/AwaitableCallback.java
+++ b/core/src/main/java/zipkin2/reporter/AwaitableCallback.java
@@ -14,7 +14,6 @@
 package zipkin2.reporter;
 
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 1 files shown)


adriancole:
oops this code used AtomicReference not volatile, but you get what I mean :P

adriancole:
added another commit. I noticed we were looping as one would in wait/notify, except that isn't relevant here. cc @anuraaga

PR #159, opened 2019-08-21
Kafka reporter

### LINKED ISSUE #158: Feature request: `KafkaReporter` in addition to `KafkaSender`
**Feature:**

Add a `KafkaReporter` to process spans individually and have the ability to use Kafka producer native batching and been able to partition spans based on `traceId`.

**Rationale**

Current `Sender` API is aimed to transfer encoded batch of spans to different back-ends; already too late to do partitioning by key, and in the case of Kafka, overlaps with producer's native batching mechanisms (`batch.size` and `linger.ms` in https://kafka.apache.org/documentation/#producerconfigs).

Instead of reusing `BoundedAsyncReporter` with `KafkaSender`, a potential `KafkaReporter` will handle the report spans itself, allowing better message definition (e.g. using `traceId` as key) and using it's own batching mechanisms.

```java
@Override public void report(Span span) {
    metrics.incrementSpans(1);
    int spanSizeInBytes = encoder.sizeInBytes(span);
    metrics.incrementSpanBytes(spanSizeInBytes);
    ProducerRecord<byte[], byte[]> record =
      new ProducerRecord<>(topic, span.traceId().getBytes(), encoder.encode(span));
    get().send(record, (metadata, exception) -> {
       if (exception != null) {
         metrics.incrementMessagesDropped(exception);
         metrics.incrementSpansDropped(1);
       }
    });
  }
```

Instead of looking for changes on the `KafkaSender` to support this improvements, `Reporter` API seems like the right abstraction level to place Kafka Producer.

**Example Scenario:**

While developing a [Kafka-based storage](https://github.com/jeqo/zipkin-storage-kafka) and a [streaming version of Zipkin Dependencies](http://github.com/sysco-middleware/zipkin-streaming-dependencies) there is a need to have spans partitioned by `traceId` in order to group them and apply aggregations that can be used to apply sampling,  dependency counts, etc.

In the Kafka Storage, we had to add a [partitioning Span Consumer](https://github.com/jeqo/zipkin-storage-kafka/blob/661ec18e7481f33bb71ff3cbc4ba6bf9fec93efe/storage/src/main/java/zipkin2/storage/kafka/KafkaSpanConsumer.java#L59-L80) to compensate this which involves an additional overhead, compared with already keyed-spans; in which case we could focus only on processing.

jeqo:
Fixes #158 

Adds a KafkaReporter to support: native Kafka batching, partitioning by `traceId`.

### DIFF
--- a/kafka/src/main/java/zipkin2/reporter/kafka/KafkaReporter.java
+++ b/kafka/src/main/java/zipkin2/reporter/kafka/KafkaReporter.java
@@ -2,6 +2,7 @@ package zipkin2.reporter.kafka;
 
 import java.io.Flushable;
 import java.io.IOException;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
@@ -239,7 +240,10 @@ public class KafkaReporter extends Component implements Reporter<Span>, Flushabl
     int spanSizeInBytes = encoder.sizeInBytes(span);
     metrics.incrementSpanBytes(spanSizeInBytes);
     ProducerRecord<byte[], byte[]> record =
-      new ProducerRecord<>(topic, span.traceId().getBytes(), encoder.encode(span));
+      new ProducerRecord<>(
+        topic,
+        span.traceId().getBytes(),
+        encoder.encodeList(Collections.singletonList(span)));
     getProducer().send(record, (metadata, exception) -> {
        if (exception != null) {
          metrics.incrementMessagesDropped(exception);
--- a/kafka/src/main/java/zipkin2/reporter/kafka/KafkaReporter.java
+++ b/kafka/src/main/java/zipkin2/reporter/kafka/KafkaReporter.java
@@ -1,0 +1,295 @@
+package zipkin2.reporter.kafka;
+
+import java.io.Flushable;
+import java.io.IOException;
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+import org.apache.kafka.clients.admin.AdminClient;
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.KafkaFuture;
+import org.apache.kafka.common.serialization.ByteArraySerializer;
+import zipkin2.CheckResult;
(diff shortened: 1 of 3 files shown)


kafka/src/main/java/zipkin2/reporter/kafka/KafkaReporter.java
```diff
@@ -0,0 +1,327 @@
...
+    if (closeCalled) throw new IllegalStateException("closed");
+    metrics.incrementSpans(1);
+    int spanSizeInBytes = encoder.sizeInBytes(span);
+    metrics.incrementSpanBytes(spanSizeInBytes);
+    ProducerRecord<byte[], byte[]> record =
+      new ProducerRecord<>(
+        topic,
+        span.traceId().getBytes(),
+        encoder.encodeList(Collections.singletonList(span)));
+    getProducer().send(record, (metadata, exception) -> {
```
adriancole: ps while not opposed to this, I also think a kafka "FinishedSpanHandler" in brave could be more effective. you could bundle messages by local root. since we don't have a field in span for this, it could only be implemented in brave.

jeqo: >  I also think a kafka "FinishedSpanHandler" in brave could be more effective. you could bundle messages by local root. 

This is something I miss from `AsyncReporter` where I can group per `traceId`, and send them together.

local root would solve the same but from brave, right?

> kafka "FinishedSpanHandler" in brave

Haven't worked with this API yet tho, but a `kafka handler` would replace `reporter` under site-specific config plus grouping spans per local root, right? 

I'm just a bit confused on the overlapping between handlers and reporter, and if it's ok to solve this on both sides or pick one.

adriancole:
main thing is handler would be limited to users of brave. otoh brave model
has local root etc. both can coexist just depends on what trying to solve.
I wouldnt have dual Kafka layers though.

just giving options for direct implementation

adriancole:
" I wouldnt have dual Kafka layers though" means I wouldnt recommend user
having to configure a handler and a reporter in a tightly coupled fashion.
we can either find a way to pass local root here (ex an extended interface
which has a traceid and local root param)

ex partitionedreporter.report(traceId, localRootId, span)

or solve somehow else

codefromthecrypt:
@jeqo this is going on 5 years old ;) go ahead and close this unless you feel like updating it to SpanHandler in brave. We can also do reporter (mostly for otel)

PR #162, opened 2019-11-11
align Sender's max msg size

### LINKED ISSUE #153: Reduce default message size
In docker testing, the default message size of 5MiB is too much.. it basically makes elasticsearch fall over. I think we should reduce significantly lower, like 1000000 bytes (kafka's size) or lower than that.

cc @anuraaga

jeqo:
Fix #153

### DIFF
--- a/activemq-client/src/main/java/zipkin2/reporter/activemq/ActiveMQSender.java
+++ b/activemq-client/src/main/java/zipkin2/reporter/activemq/ActiveMQSender.java
@@ -76,7 +76,7 @@ public final class ActiveMQSender extends Sender {
     ActiveMQConnectionFactory connectionFactory;
     String queue = "zipkin";
     Encoding encoding = Encoding.JSON;
-    int messageMaxBytes = 100000; // arbitrary to match kafka, messages theoretically can be 2GiB
+    int messageMaxBytes = 500_000; // arbitrary to match kafka, messages theoretically can be 2GiB
 
     public Builder connectionFactory(ActiveMQConnectionFactory connectionFactory) {
       if (connectionFactory == null) throw new NullPointerException("connectionFactory == null");
@@ -102,7 +102,7 @@ public final class ActiveMQSender extends Sender {
       return this;
     }
 
-    /** Maximum size of a message. Default 1000000. */
+    /** Maximum size of a message. Default 500000. */
     public Builder messageMaxBytes(int messageMaxBytes) {
       this.messageMaxBytes = messageMaxBytes;
       return this;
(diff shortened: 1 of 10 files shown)


activemq-client/src/main/java/zipkin2/reporter/activemq/ActiveMQSender.java
```diff
@@ -76,7 +76,7 @@ public static Builder newBuilder() {
     ActiveMQConnectionFactory connectionFactory;
     String queue = "zipkin";
     Encoding encoding = Encoding.JSON;
-    int messageMaxBytes = 100000; // arbitrary to match kafka, messages theoretically can be 2GiB
+    int messageMaxBytes = 500_000; // arbitrary to match kafka, messages theoretically can be 2GiB
```
anuraaga: If I'm not mistaken, this doesn't match Kafka anymore which is 1MB. Or maybe there's something related to bi-directionality here.

Could you explain more why we want to use half of Kafka max size as our default? If the reason is just to be somewhat conservative as a default and let users optimize themselves, it makes sense, but in that case I guess we don't need to mention Kafka in these comments.

jeqo: @anuraaga yes, it's a conservative default as you mention. I will improve docs.

activemq-client/src/main/java/zipkin2/reporter/activemq/ActiveMQSender.java
```diff
@@ -102,7 +102,7 @@ public Builder encoding(Encoding encoding) {
       return this;
     }
 
-    /** Maximum size of a message. Default 1000000. */
+    /** Maximum size of a message. Default 500KB. */
     public Builder messageMaxBytes(int messageMaxBytes) {
```
anuraaga: Do you think we should update these builder methods to validate the size is less than the implementation limit (e.g., 1MB for kafka, 2GB for activemq)?

core/src/main/java/zipkin2/reporter/Sender.java
```diff
@@ -51,6 +51,9 @@
   /**
    * Maximum bytes sendable per message including overhead. This can be calculated using {@link
    * #messageSizeInBytes(List)}
+   *
+   * Defaults to 500_000 bytes as conservative default given that Kafka default max.request.size is
```
anuraaga: Don't think we need to tie to Kafka here either. How about this?

`Defaults to 500KB as a conservative default. You may get better or reduced performance by changing this value based on, e.g., machine size or network bandwidth in your infrastructure. Finding a perfect value will require trying out different values in production, but the default should work well enough in most cases.`

activemq-client/src/main/java/zipkin2/reporter/activemq/ActiveMQSender.java
jeqo: not sure, as this limit can be customized (in Kafka at least)

jeqo:
thanks @anuraaga for the review! I'll go ahead and merge this one.

PR #165, opened 2019-11-29
Fix race condition in AsyncReporter

### LINKED ISSUE #164: Race condition flushing spans in AsyncReporter
Running zipkin-reporter 2.3.1, there is about a 1:100,000 chance we get the following Exception in our logs:
```
AsyncReporter{URLConnectionSender{http://zipkin:9411/api/v2/spans}}had an unhandled exception
java.lang.IllegalStateException: closed
	at zipkin2.reporter.AsyncReporter$BoundedAsyncReporter.flush(AsyncReporter.java:241)
	at zipkin2.reporter.AsyncReporter$Builder$1.run(AsyncReporter.java:179)
```

We see it somewhat frequently because we run hundreds of thousands of "micro tasks" a day.  It looks like it is a fairly simple race condition in AsyncReporter that is causing this.

Ref: [AsyncReporter 2.3.1](https://github.com/openzipkin/zipkin-reporter-java/blob/2.3.1/core/src/main/java/zipkin2/reporter/AsyncReporter.java#L1780).  The most recent version of the code has the same issue just relocated: [AsyncReporter 2.11.1](https://github.com/openzipkin/zipkin-reporter-java/blob/2.11.1/core/src/main/java/zipkin2/reporter/AsyncReporter.java#L353)

Near as I can tell, the workflow (relative to 2.11.1) is:
1. `run()` checks to see if the reporter is closed
2. It is not closed so we move on and attempt to `flush()`
3. A separate Thread closes the reporter
4. The `flush()` method fails because we're now closed

Again, this is super rare as you really have to "thread the needle" but it also looks easy enough to fix.  Will submit pull request shortly.

Logic-32:
Fixes #164
Moved "closed" check up to public method so that internal operations can check the closed status on their own and flush accordingly without risk of running into exceptions.
Moved initialization of closed variable to prevent implicit creation of second constructor.

### DIFF
--- a/core/src/main/java/zipkin2/reporter/AsyncReporter.java
+++ b/core/src/main/java/zipkin2/reporter/AsyncReporter.java
@@ -207,7 +207,7 @@ public abstract class AsyncReporter<S> extends Component implements Reporter<S>,
 
   static final class BoundedAsyncReporter<S> extends AsyncReporter<S> {
     static final Logger logger = Logger.getLogger(BoundedAsyncReporter.class.getName());
-    final AtomicBoolean closed = new AtomicBoolean(false);
+    final AtomicBoolean closed;
     final BytesEncoder<S> encoder;
     final ByteBoundedQueue<S> pending;
     final Sender sender;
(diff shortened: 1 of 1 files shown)


Logic-32:
FWIW, I really wanted to update this to use a ScheduledExecutorService and do a more involved change.  But after digging in a bit I realize that it looks like the intent here is to block closing until the async flushing is complete.  As such, it seemed more reasonable to me to just remove the double-check of `closed` along the async flushing line.

PR #166, opened 2019-12-04
Use special exception to mark a closed reporter to not conflict with …

anuraaga:
…other libraries.

A sender may throw `IllegalStateException` for reasons other than being closed as it's a standard Java exception. So if we want to handle it specially, we should use a specific type for it. As there could be usage of the reporter with senders outside our repos or older versions, a less precise check is still kept as well.

/cc @elefeint

Fixes https://github.com/openzipkin/zipkin-gcp/issues/127

### DIFF
--- a/core/src/main/java/zipkin2/reporter/AsyncReporter.java
+++ b/core/src/main/java/zipkin2/reporter/AsyncReporter.java
@@ -13,8 +13,11 @@
  */
 package zipkin2.reporter;
 
+import static java.lang.String.format;
+import static java.util.logging.Level.FINE;
+import static java.util.logging.Level.WARNING;
+
 import java.io.Flushable;
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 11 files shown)


core/src/main/java/zipkin2/reporter/AsyncReporter.java
```diff
@@ -282,7 +282,7 @@ void flush(BufferNextMessage<S> bundler) {
 
       try {
         sender.sendSpans(nextMessage).execute();
-      } catch (IOException | RuntimeException | Error t) {
```
shakuzen: I'm wondering why we had this level of specificity before. What types of exceptions did we not want to catch and is it okay to catch them now?

anuraaga: The specificity hasn't changed - the previous signature caught all unchecked exceptions, and the only possible checked exception `IOException`. But I find this to be less readable than just going with `Throwable` which is semantically perfect (any thrown exception) and always less code than explicitly listing checked exceptions.

Sorry for adding this cleanup without pointing it out it's admittedly unrelated to the issue.

PR #171, opened 2020-04-20
Adds Brave integration

adriancole:
This takes a copy of `ZipkinSpanHandler` in Brave so that we can
eventually decouple Brave from a compile depedency on Zipkin Reporter.

This also allows the commonly requested feature (such as exists in Sleuth)
to allow multiple simultaneous reporters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,10 @@ the `Reporter<Span>` here to a Brave `FinishedSpanHandler`.
 
 Ex.
 ```java
- tracingBuilder.addFinishedSpanHandler(ZipkinSpanHandler.create(reporter));
+tracingBuilder.addFinishedSpanHandler(ZipkinSpanHandler.create(reporter));
 ```
+
+See [zipkin-reporter-brave](brave/README.md) for more details.
 
 ## Spring Beans
 If you are trying to trace legacy applications, you may be interested in
--- a/README.md
+++ b/README.md
@@ -39,8 +39,12 @@ reporter.report(span);
 
 ## Brave
 Those using the [Brave library](https://github.com/openzipkin/brave) can adapt
-the `Reporter<Span>` here to a Brave
+the `Reporter<Span>` here to a Brave.
 
+Ex.
+```java
+ tracingBuilder.addFinishedSpanHandler(ZipkinFinishedSpanHandler.create(reporter));
+```
 
 ## Spring Beans
 If you are trying to trace legacy applications, you may be interested in
--- a/README.md
+++ b/README.md
@@ -39,11 +39,11 @@ reporter.report(span);
 
 ## Brave
 Those using the [Brave library](https://github.com/openzipkin/brave) can adapt
-the `Reporter<Span>` here to a Brave.
+the `Reporter<Span>` here to a Brave `FinishedSpanHandler`.
 
 Ex.
 ```java
- tracingBuilder.addFinishedSpanHandler(ZipkinFinishedSpanHandler.create(reporter));
+ tracingBuilder.addFinishedSpanHandler(ZipkinSpanHandler.create(reporter));
 ```
 
 ## Spring Beans
--- a/README.md
+++ b/README.md
@@ -37,6 +37,11 @@ reporter = AsyncReporter.create(URLConnectionSender.create("http://localhost:941
 reporter.report(span);
 ```
 
+## Brave
+Those using the [Brave library](https://github.com/openzipkin/brave) can adapt
+the `Reporter<Span>` here to a Brave
+
+
 ## Spring Beans
 If you are trying to trace legacy applications, you may be interested in
 [Spring XML Configuration](spring-beans/). This allows you to trace legacy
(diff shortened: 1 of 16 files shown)


adriancole:
going to add spring-beans support next, so that there's no awkward reverse dependency

README.md
```diff
@@ -37,6 +37,15 @@ reporter = AsyncReporter.create(URLConnectionSender.create("http://localhost:941
 reporter.report(span);
 ```
 
+## Brave
+Those using the [Brave library](https://github.com/openzipkin/brave) can adapt
+the `Reporter<Span>` here to a Brave.
```
anuraaga, line 42: to a Brave `FiinishedSpanHandler`.

brave/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>zipkin-reporter</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.zipkin.brave</groupId>
+      <artifactId>brave</artifactId>
+      <!-- Don't pin Brave -->
```
anuraaga, line 47: Don't we need to at least define some min Brave version though?

brave/src/main/java/zipkin2/reporter/brave/ZipkinSpanHandler.java
```diff
@@ -0,0 +1,203 @@
...
+      .id(context.spanId())
+      .name(span.name());
+
+    long start = span.startTimestamp(), finish = span.finishTimestamp();
+    result.timestamp(start);
+    if (start != 0 && finish != 0L) result.duration(Math.max(finish - start, 1));
+
+    // use ordinal comparison to defend against version skew
+    brave.Span.Kind kind = span.kind();
+    if (kind != null && kind.ordinal() < Span.Kind.values().length) {
```
anuraaga, line 137: Just wondering do we have a unit test that checks for ordinal drift?

brave/pom.xml
adriancole: not worth it, imho and we don't do the same anywhere else either, not in brave JMS for example, or any other project.

if we do compile scope vs provided here, it is more likely to end up as a classpath clash in our gitter channel and I really don't want to do that.

adriancole: ps the caution in general is about Nteen build tools and some dimensions of all of them and how they interpret Maven variable version range. I personally can't take on that risk especially knowing we knocked all things like this provided to dig out of problems in the first place. Not something I want to waranty, and likely it would be me.

this treats brave version conventionally like literally everything else saying the primary library owns that version. The only difference is I put a comment to that nature, where in other poms it is implied.

README.md
adriancole: sg

brave/src/main/java/zipkin2/reporter/brave/ZipkinSpanHandler.java
adriancole: doubt it, and hope we never add any kinds, but this does deserve a unit test. it would have to be quite a thing (like powermock?) to change an enum's ordinal to make this fail.

anuraaga: Was just thinking of one that checks that the enum items with same name have same ordinal. But true kind change seems unlikely enough that not a big deal

adriancole: probably a switch statement is easier than powermock.

This was originally here to allow producer/consumer kind which didn't exist in one or the other at the time (I don't recall). I don't expect us to add any more kinds and someone can do powermock later if we do :)

adriancole: I changed to a static initialized map, as I think it is overall easier to comprehend and  cleaner. thanks for noticing

adriancole:
@anuraaga I tested this with spring 2.5 XML both earliest Brave (5.6) and snapshot (which gets rid of the log messages). PTAL?

adriancole:
nap was useful. the generic type of the asyncreporter is wrong! I will fix it in a bit

adriancole:
nah can't do what I want right now: `AsyncReporter<MutableSpan>`. the MutableSpan type doesn't have the fields we need yet to size up the span (traceID, parentId etc). Will revisit once Brave 5.12 is out.

adriancole:
I'm currently of the opinion that setting minimum brave version to 5.12 is worth it as it avoids a lot of hacks. Folks can safely move from 5.6 to 5.12 anyway and worst case we have to teach them how to use the bom. By setting minimum to 5.12, we can actually skip the step of converting to zipkin span first, and directly encode into v2 json or proto.

anuraaga:
This library basically forked off of brave in some sense - so making the minimum the point of fork seems fine to me.

adriancole:
we have an interdependency on https://github.com/openzipkin/brave/pull/1167 so I'm merging and cutting zipkin-reporter with a pending version of brave. It will eventually resolve later when brave is cut.

adriancole:
thanks for the eyes @anuraaga and @jeqo !

PR #174, opened 2020-05-13
Adds AsyncZipkinSpanHandler to write Brave's span directly to JSON

adriancole:
Formerly, MutableSpan needs to convert first to zipkin2.Span before
sending to Zipkin. Now, provided the format is Zipkin V2, this can be
done directly, with no conversion overhead.

Note: the conversion overhead is inline on the production request, so
the practice here is preferred:

```java
sender = URLConnectionSender.create("http://localhost:9411/api/v2/spans");
zipkinSpanHandler = AsyncZipkinSpanHandler.create(sender); // don't forget to close!
tracingBuilder.addSpanHandler(zipkinSpanHandler);
```

If you have to use a different format, though, you can still use
conversion. Ex.
```java
reporter = AsyncReporter.builder(URLConnectionSender.create("http://localhost:9411/api/v1/spans"))
                        .build(SpanBytesEncoder.JSON_V1);
tracingBuilder.addSpanHandler(ZipkinSpanHandler.create(reporter));
```

Depends on https://github.com/openzipkin/brave/pull/1212

### DIFF
--- a/brave/src/main/java/zipkin2/reporter/brave/AsyncZipkinSpanHandler.java
+++ b/brave/src/main/java/zipkin2/reporter/brave/AsyncZipkinSpanHandler.java
@@ -1,0 +1,153 @@
+/*
+ * Copyright 2016-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.reporter.brave;
+
+import brave.Tag;
+import brave.handler.MutableSpan;
+import brave.handler.SpanHandler;
+import java.io.Closeable;
+import java.io.Flushable;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import zipkin2.reporter.AsyncReporter;
+import zipkin2.reporter.ReporterMetrics;
+import zipkin2.reporter.Sender;
+
+/**
+ * A {@link brave.handler.SpanHandler} that queues spans on {@link #end} to bundle and send as a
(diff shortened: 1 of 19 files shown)


adriancole:
@jeqo @anuraaga finally the holy grail.. brave directly to reporting messages!

adriancole:
I'm going to test with mvc example XML tomorrow morning

adriancole:
also will switch to encoder instead of fussing with the builder.. like this.
```java
spanReporter = AsyncReporter.builder(URLConnectionSender.create("http://localhost:9411/api/v2/spans"));
                       .build(BraveEncoder.JSON_V2);
tracingBuilder.addSpanHandler(ZipkinSpanHandler.create(spanReporter));
```

adriancole:
@anuraaga PTAL I think the design is right now.. I'll test it through things like dubbo though

brave/src/main/java/zipkin2/reporter/brave/AsyncZipkinSpanHandler.java
```diff
@@ -0,0 +1,154 @@
...
+import java.io.Closeable;
+import java.io.Flushable;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import zipkin2.reporter.AsyncReporter;
+import zipkin2.reporter.ReporterMetrics;
+import zipkin2.reporter.Sender;
+
+/**
+ * A {@link brave.handler.SpanHandler} that queues spans on {@linkplain #end end} to send in bulk
```
anuraaga, line 28: Believe `{@linkplain #end end}` is equivalent to `{@link #end}`

brave/src/main/java/zipkin2/reporter/brave/AsyncZipkinSpanHandler.java
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * @see ZipkinSpanHandler if you need to use a different format
+ * @see brave.Tracing.Builder#addSpanHandler(SpanHandler)
+ * @since 2.14
+ */
+public final class AsyncZipkinSpanHandler extends ZipkinSpanHandler
+    implements Closeable, Flushable {
+  /** @since 2.14 */
+  public static AsyncZipkinSpanHandler create(Sender sender) {
+    return newBuilder(sender).build(); // can't be no-op
```
anuraaga, line 48: FWIW I don't understand the comment - maybe can just remove it.

adriancole:
thanks for the look @anuraaga

PR #175, opened 2020-05-17
Allows ZipkinSpanHandler to be reconfigured for "firehose mode"

adriancole:
The routine configuration of secondary sampling is to set
`alwaysReportSpans` so that the same zipkin reporter configured will now
be used also for secondary reporting.

We deprecated `Tracing.Builder.alwaysReportSpans()`, but before this
change, there was no viable alternative, as the sampling component
doesn't control the span reporting one.

This solves the problem by adding `ZipkinSpanHandler.toBuilder()`, which
can be called after the zipkin handler is added to `Tracing.Builder`.

To reduce burden, this also makes `AsyncReporter` lazy start its flusher
which is something it probably should have always done.

See https://github.com/openzipkin-contrib/zipkin-secondary-sampling/pull/33

### DIFF
--- a/brave/src/main/java/zipkin2/reporter/brave/ZipkinSpanHandler.java
+++ b/brave/src/main/java/zipkin2/reporter/brave/ZipkinSpanHandler.java
@@ -18,8 +18,10 @@ import brave.Tags;
 import brave.handler.MutableSpan;
 import brave.handler.SpanHandler;
 import brave.propagation.TraceContext;
+import java.io.Closeable;
 import zipkin2.Span;
 import zipkin2.codec.SpanBytesEncoder;
+import zipkin2.reporter.AsyncReporter;
 import zipkin2.reporter.Reporter;
 
 /**
@@ -43,7 +45,7 @@ import zipkin2.reporter.Reporter;
  * @see brave.Tracing.Builder#addSpanHandler(SpanHandler)
  * @since 2.13
  */
-public class ZipkinSpanHandler extends SpanHandler {
+public class ZipkinSpanHandler extends SpanHandler implements Closeable {
   /** @since 2.13 */
   // SpanHandler not ZipkinSpanHandler as it can coerce to NOOP
   public static SpanHandler create(Reporter<Span> spanReporter) {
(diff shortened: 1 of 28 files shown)


brave/src/main/java/zipkin2/reporter/brave/ZipkinSpanHandler.java
```diff
@@ -53,28 +55,40 @@ public static SpanHandler create(Reporter<Span> spanReporter) {
...
-      return new ZipkinSpanHandler(new ConvertingSpanReporter(spanReporter, errorTag),
-          alwaysReportSpans);
-    }
+  /**
+   * Implementations that throw exceptions on close have bugs. This may result in log warnings,
+   * though.
+   *
+   * @since 2.15
+   */
+  @Override public void close() {
```
adriancole, line 80: without this, the caller will have to play games to figure out if it should be closed or not. This lets us get the wildcard "Closeable" like we do with zipkin2.Component, but without having to couple this type to zipkin2 package.

core/bnd.bnd
```diff
@@ -1,4 +1,5 @@
 Import-Package: \
   *
 Export-Package: \
-  zipkin2.reporter
+  zipkin2.reporter,\
+  brave.internal;braveinternal=true;mandatory:=braveinternal
```
adriancole, line 5:
```suggestion
  zipkin2.reporter.internal;zipkinreporterinternal=true;mandatory:=zipkinreporterinternal
```

brave/src/main/java/zipkin2/reporter/brave/ZipkinSpanHandler.java
```diff
@@ -53,28 +55,40 @@ public static SpanHandler create(Reporter<Span> spanReporter) {
...
-  static final class ConvertingSpanReporterBuilder extends Builder {
-    final Reporter<Span> spanReporter;
-
-    ConvertingSpanReporterBuilder(Reporter<Span> spanReporter) {
-      this.spanReporter = spanReporter;
-    }
+  /**
+   * Allows this instance to be reconfigured, for example {@link Builder#alwaysReportSpans(boolean)}.
+   *
+   * <p><em>Note:</em>If the instance  {@link AsyncReporter#close()} if you no longer need this
```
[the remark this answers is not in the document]
adriancole: or mumbling :D

adriancole:
thanks for having a look @anuraaga

PR #176, opened 2020-05-26
Actually run RabbitMQ integration tests

adriancole:
Before, these only ran locally. This makes them use Docker and never
local setup. The fallback is not that helpful, and makes interpreting
failures complex.

Similar to https://github.com/openzipkin/zipkin/issues/1748

### DIFF
--- a/amqp-client/src/test/java/zipkin2/reporter/amqp/RabbitMQSenderRule.java
+++ b/amqp-client/src/test/java/zipkin2/reporter/amqp/RabbitMQSenderRule.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2016-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.reporter.amqp;
+
+import java.time.Duration;
+import org.junit.AssumptionViolatedException;
+import org.junit.ClassRule;
+import org.junit.rules.ExternalResource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.Container.ExecResult;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import zipkin2.CheckResult;
+
+import static zipkin2.Call.propagateIfFatal;
+
+/** This should be used as a {@link ClassRule} as it takes a very long time to start-up. */
+class RabbitMQSenderRule extends ExternalResource {
+  static final Logger LOGGER = LoggerFactory.getLogger(RabbitMQSenderRule.class);
+  static final String IMAGE = "rabbitmq:3.7-management-alpine";
+  static final String QUEUE = "zipkin-test1";
+  static final int RABBIT_PORT = 5672;
+
+  static final class RabbitMQContainer extends GenericContainer<RabbitMQContainer> {
+    RabbitMQContainer(String image) {
+      super(image);
+      addExposedPorts(RABBIT_PORT);
+      this.waitStrategy = Wait.forLogMessage(".*Server startup complete.*", 1)
+          .withStartupTimeout(Duration.ofSeconds(60));
+    }
+  }
+
+  RabbitMQContainer container;
+
+  @Override protected void before() {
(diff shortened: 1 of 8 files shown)


amqp-client/src/test/java/zipkin2/reporter/amqp/RabbitMQSenderRule.java
```diff
@@ -0,0 +1,111 @@
...
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import zipkin2.CheckResult;
+
+import static zipkin2.Call.propagateIfFatal;
+
+/** This should be used as a {@link ClassRule} as it takes a very long time to start-up. */
+class RabbitMQSenderRule extends ExternalResource {
+  static final Logger LOGGER = LoggerFactory.getLogger(RabbitMQSenderRule.class);
+  static final String IMAGE = "rabbitmq:3.7-management-alpine";
```
adriancole, line 32: this image has rabbitmqadmin which is a lot more reliable than the java variants in my experience trying to get anything to pass without lots of sleeps.

amqp-client/src/test/java/zipkin2/reporter/amqp/RabbitMQSenderRule.java
```diff
@@ -0,0 +1,111 @@
...
+  static final String QUEUE = "zipkin-test1";
+  static final int RABBIT_PORT = 5672;
+
+  static final class RabbitMQContainer extends GenericContainer<RabbitMQContainer> {
+    RabbitMQContainer(String image) {
+      super(image);
+      addExposedPorts(RABBIT_PORT);
+      this.waitStrategy = Wait.forLogMessage(".*Server startup complete.*", 1)
```
adriancole, line 40: hint was from here https://github.com/testcontainers/testcontainers-java/blob/master/modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java

we don't want to use the upstream container as we'd have to wrap it anyway as it doesn't check things like status of commands, and we can't do "docker.skip" without wrapping it again as far as I know either..

finally I found almost no usage of that container on github so there are probably glitches we don't want to inherit.

amqp-client/src/test/java/zipkin2/reporter/amqp/RabbitMQSenderTest.java
```diff
@@ -0,0 +1,57 @@
...
+import zipkin2.reporter.ClosedSenderException;
+import zipkin2.reporter.Sender;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static zipkin2.TestObjects.CLIENT_SPAN;
+import static zipkin2.reporter.amqp.ITRabbitMQSender.send;
+
+/** This works against a running RabbitMQ server on localhost */
+public class RabbitMQSenderTest {
```
adriancole, line 28: I pulled tests that don't need a container or anything into a unit test

amqp-client/src/test/java/zipkin2/reporter/amqp/RabbitMQSenderRule.java
bsideup: We would, however, love to hear your feedback on the official module :) Maybe as a follow up

Also, I know a few usages outside of GitHub (and, IIRC, one of them is from Blizzard). Plus:
- https://github.com/xing/java-beetle/blob/98302c1c0bf418040883ead6a542553330a1a533/beetle-core/pom.xml#L53
- https://github.com/flowable/flowable-engine/blob/78435d430795424f20357ba2f24ac88e9bfd585e/modules/flowable-event-registry-spring/pom.xml#L204
- https://github.com/Activiti/activiti-cloud/blob/00b0c68351d3c1ece5f88478c5a536c4243bbf9b/activiti-cloud-audit-service/activiti-cloud-starter-audit/pom.xml#L155
- https://github.com/hazelcast/hazelcast-jet/blob/72904b66efc6e3e9e2f9d327daa17cc7ccae4cfc/hazelcast-jet-core/pom.xml#L152

amqp-client/src/test/java/zipkin2/reporter/amqp/RabbitMQSenderRule.java
```diff
@@ -0,0 +1,109 @@
...
+  RabbitMQContainer container;
+
+  @Override protected void before() {
+    if ("true".equals(System.getProperty("docker.skip"))) {
+      throw new AssumptionViolatedException("Skipping startup of docker " + IMAGE);
+    }
+
+    try {
+      LOGGER.info("Starting docker image " + IMAGE);
+      container = new RabbitMQContainer(IMAGE);
```
bsideup, line 54: FYI we no longer perform _any_ side effect in constructors, so it is safe to make it final

amqp-client/src/test/java/zipkin2/reporter/amqp/RabbitMQSenderRule.java
adriancole: I was looking less at poms and more at source files.. didn't see much of anything using the extended commands. For sure on follow-up though! works for me

PR #179, opened 2020-08-21
Bumps versions and avoids JDK 14 OSGi resolution problem

adriancole:
@reta can you verify this?

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -29,7 +29,7 @@
 
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
-    <jmh.version>1.23</jmh.version>
+    <jmh.version>1.25</jmh.version>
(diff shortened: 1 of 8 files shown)


benchmarks/pom.xml
```diff
@@ -29,7 +29,7 @@
 
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
-    <jmh.version>1.23</jmh.version>
+    <jmh.version>1.25</jmh.version>
```
anuraaga, line 32: Nice https://hg.openjdk.java.net/code-tools/jmh/rev/fa4049d004a4 https://hg.openjdk.java.net/code-tools/jmh/rev/50b5cdf9c528

reta:
@adriancole the MANIFEST.MF is good, CXF build passes against the snapshots with no issues, thanks!

PR #180, opened 2020-09-10
Adds b3:0 header to avoid trace amplification when using proxies

adriancole:
Same as https://github.com/openzipkin/zipkin-go/pull/176


shakuzen (commented):
LGTM

anuraaga (commented):
Just wondering, shouldn't we set `x-b3` too? Aren't there proxies that don't support `b3` and would have write amplification? Or does `proxy` == `envoy` now adays (I'd buy it :P)

adriancole:
good question: more bytes on the wire, and would prefer to let such an old proxy that doesn't know "b3" yet, and needs this, become known! I'd prefer to convince them to support "b3" rather than force another and longer header on everyone.

adriancole:
https://github.com/openzipkin/b3-propagation/pull/42

PR #195, opened 2020-12-15
Upgrade okhttp major version to 4

usulkies:


adriancole:
thanks for the interest! I'm curious how the version update impacts you. Right now, we set "provided" so we should expect to not interfere with those using 4.x .. is that not working? One thing in tracing is we can't expect everyone to be on the same version, so it is more a range issue than a singular issue one.

usulkies:
> Right now, we set "provided" so we should expect to not interfere with those using 4.x .. is that not working? 

Where is this "provided"?
In the okhttp3 module there is this dependency
image
Which is a direct dependency on okhttp.
Since we are dependant on zipkin-sender-okhttp3, it conflicts with other dependencies which are already using the OkHttp4.

Anyway, now I see that the artifact name is zipkin-sender-okhttp3, which should be changed to zipkin-sender-okhttp4, if this change is to get in.

usulkies:
I can exclude okhttp3 from the zipkin-sender-okhttp3 dependency, but I thought to come to aid and fix this for the community.

adriancole:
My mistake. in brave we set it to provided as people explicitly already choose a version. In reporter we don't. I think there was a discussion about this in the past.

Note that for some this is a fix and for others, it won't be. okhttp v4 brings in a kotlin dependency, so is substantially larger that 3. It might still be ok to flip the default, but would need 👍 from a few users to justify considering this.

adriancole:
random selection for spam: @beckje01 @reta @OlegDokuka @jonathan-lo @beiwei30 @yschimke if any of you prefer we flip the default zipkin reporter to okhttp4 instead of 3, please 👍 this PR. If you don't want it, please comment when it would be appropriate. Note, this is still binary compatible regardless, it is mainly about the implicit dependency.

If there aren't any adverse notes, I'm keen to merge this as it has been quite a while and pros and cons are understood, or understandable at this point.

adriancole:
If we move forward, on this, I sign myself up for release notes. I would basically include the implicits here. Ex.

3.12.12 was the last version to support Java 1.7.

We've already overrun that 3.14.9 -> 4.9.0, so it isn't a matter of changing floor Java version... that remains the same: Java 1.8 unless you override the dependency.

Basically, this is solely a matter of the new line of okhttp or not.

We may want to run back around after this to make an invoker test as we have in brave, to prove that the older versions still work. https://github.com/openzipkin/brave/blob/master/instrumentation/okhttp3/src/it/okhttp3_v3/README.md

notably, this might require making unit tests a little more fancy, or cherry-picking which IT is run for the older version.

yschimke:
If you have already moved past 3.12.12. Keep going.

That's said are you happy to start accidentally building on new APIs?

adriancole:
> That's said are you happy to start accidentally building on new APIs?

an invoker test or similar will need to be there to prevent conflicts on that, yep. It would be easy to break meanwhile even if this change doesn't break api. If we wanted to use the new apis (and this particular codebase doesn't do much that would need one), we could make a platform type detection to allow coexistence I think.

reta:
@adriancole no objections to move to 4.x, if we could keep 3.12.x as minimal in OSGi manifest [3.12.0, 5.0) would be great, I would be happy to give it a try with snapshots, wdyt?

adriancole:
@reta good idea on the manifest thing. sure will appreciate it. Will merge this tomorrow unless screams happen before.

adriancole:
@reta you mind starting by fixing (or looking at) our brave osgi manifest for brave-instrumentation-okhttp3 meanwhile? I have a suspicion it is wrong even if our tests cover the range properly. Once in we can sortof mechanically do the same here.

reta:
@adriancole sure thing

adriancole:
this drifted due to things including a rebase. I don't plan to personally implement the invoker test that would be required to merge this. I'm stopping reworking pull requests to do the hard work. I'm also a volunteer and have no income either.

someone who wants to finish this can look at this and do similar here to make sure we don't accidentally drift on v3  https://github.com/openzipkin/brave/tree/master/instrumentation/okhttp3/src/it/okhttp3_v3

PR #200, opened 2021-02-26
allow passing client to sender builder

arun0009:
Allow passing custom OkHttpClient to OkHttpSender builder so we can configure a proxy, httpLoggingInterceptor, etc on OkHttpClient.


jorgheymans:
Is there an overlap with #195 here ? How would an upgrade to okhttp4 impact this ? It seems they have gone to great length to stay compatible but there are some differences. https://square.github.io/okhttp/upgrading_to_okhttp_4/

arun0009:
It's not an overlap, with current implementation irrelevant of OkHttp version, I can't configure OkHttpClient with for e.g. proxy etc.

arun0009:
@anuraaga - can you please review?

PR #210, opened 2023-04-16
chore: update kafka to 3.4

jeqo:


kafka/src/main/java/zipkin2/reporter/kafka/KafkaSender.java
```diff
@@ -306,7 +306,7 @@ AdminClient getAdminClient() {
     if (producer != null) producer.close();
     AdminClient adminClient = this.adminClient;
     // Intentionally using deprecated method to avoid pinning newer Kafka client
-    if (adminClient != null) adminClient.close(1, TimeUnit.SECONDS);
+    if (adminClient != null) adminClient.close();
```
jeqo, line 309: Update admin close to undefined timeout as method with timeunit is not available anymore.

shakuzen: What's the minimum Kafka version supported now? We should probably document this somewhere.

jeqo: With this change we fallback to a close method that is supported across versions -- if we were to use `.close(Duration)`, then it will limit versions; also sniffer was complaining about Duration as it's JDK 8 AFAIR.

shakuzen: I see. So we've kept library version compatibility, but now the behavior is changed - it will wait indefinitely for close now. I guess that's not such a big deal since we already wait indefinitely to close the `KafkaProducer`. Thanks for the explanation!

shakuzen (approved):

kafka/src/main/java/zipkin2/reporter/kafka/KafkaSender.java
shakuzen: I guess the comment on line 308 can be removed now, though, since we're no longer using a deprecated method.

PR #212, opened 2023-04-19
spring-beans 2.5.6 -> 3.0.0.RELEASE; fix @Bean AutoConfiguration mult…

victory-wu:
After Springbean version 3.0, FactoryBeans automatically configured through @ bean encountered a not found bean error

The reason is that no specific generic type was specified。

In order to reduce the impact, I only upgraded to the minimum version number that supports AbstractFactoryBean<T>

The following is a code example。

erro：not fund   AsyncReporter  Sender 。。。

``` java

    @Bean
    public  OkHttpSenderFactoryBean sender() {
        OkHttpSenderFactoryBean okHttpSenderFactoryBean = new OkHttpSenderFactoryBean();
        okHttpSenderFactoryBean.setEndpoint("http://localhost:9412/api/v2/spans");
        return okHttpSenderFactoryBean;
    }

    @Bean
    public  AsyncReporterFactoryBean reporter(Sender sender) throws Exception {
        AsyncReporterFactoryBean asyncReporterFactoryBean = new AsyncReporterFactoryBean();
        asyncReporterFactoryBean.setSender(sender);
        asyncReporterFactoryBean.setCloseTimeout(500);
        return asyncReporterFactoryBean;
    }

    @Bean
    public ZipkinSpanHandlerFactoryBean spanHandler(AsyncReporter reporter) {
        ZipkinSpanHandlerFactoryBean factoryBean = new ZipkinSpanHandlerFactoryBean();
        factoryBean.setSpanReporter(reporter);
        return factoryBean;
    }

```

### DIFF
--- a/spring-beans/pom.xml
+++ b/spring-beans/pom.xml
@@ -101,7 +101,7 @@
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-beans</artifactId>
-      <version>2.5.6</version>
+      <version>3.0.0.RELEASE</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
--- a/spring-beans/src/main/java/zipkin2/reporter/beans/ActiveMQSenderFactoryBean.java
+++ b/spring-beans/src/main/java/zipkin2/reporter/beans/ActiveMQSenderFactoryBean.java
@@ -19,7 +19,7 @@ import zipkin2.codec.Encoding;
 import zipkin2.reporter.activemq.ActiveMQSender;
 
 /** Spring XML config does not support chained builders. This converts accordingly */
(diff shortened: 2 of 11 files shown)


codefromthecrypt:
hmm I can understand what you are asking about. On the other hand, we do have code that uses later versions of spring with this dependency. Also, this dependency is intentionally old to support 2.5. To update this will make those apps stop working. Can you have another look, perhaps at the brave-example to see if there is another way to accomplish your goal? https://github.com/openzipkin/brave-example/blob/7c3d4ce09ab81066ae909273792f19cee34229ef/webmvc4-boot/src/main/java/brave/example/TracingAutoConfiguration.java

codefromthecrypt:
I haven't had time to create tests for this as there is some more urgent work I have to complete. If someone else can, it is appreciated.

PR #213, opened 2023-08-25
OTLP handler

marcingrzejszczak:
First pass on seeing how difficult it would be to add OTLP support with Brave. 

I was naive to think you can take a `MutableSpan`, convert it to proto version of whatever OTLP accepts and then use Protobuff's JSON conversion. It turns out that OTLP has deviated away from what Protobuf does out of the box - https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#json-protobuf-encoding.

Things to look at:

- I needed use tools like Jayway's JSON Path to convert ids from Base64 or convert text into longs
- I have just one test that tests nothing really
- I use a URLConnection for that which is not a good idea.
- I made a `Sender` allow to send just bytes not a list of bytes cause OTLP doesn't expect a list. With that change we can reuse all senders we have (with slight changes)
- There's no async support, nor batching

Other than that the result looks promising. I don't understand why the annotation has a negative time but I will get to the bottom of that next week.

jaeger


shakuzen (commented):
Thank you for your perseverance through all the iterations on this. I think it landed somewhere nice both on maintenance and usage. I left some minor comments, but overall this looks good to me. I did wonder a bit about the naming of the module. In a theoretical future where we tried to support OTLP in JSON format or OTLP over gRPC, what would we do? The module naming now kind of implies it would all go in the same module. Would that be feasible? I know it's a bikeshed issue, but once we release the module with whatever naming, it will be semi-permanent.

shakuzen:
I added a few more reviewers. I'd like to get input from at least one other Zipkin member on this.

anuraaga:
Randomly saw the question about JSON / gRPC. Just as a reference, otel sdk has just one artifact, mainly due to reimplementing the gRPC wire protocol on top of okhttp and directly marshaling without protobuf (heavily inspired by brave's proto serialization). I described it in a post once here

https://github.com/open-telemetry/opentelemetry-specification/discussions/1996

It's quite a bit of work so not something to think about for a first version but maybe that can help if deciding on artifacts in the future.

codefromthecrypt:
made a [comment](https://github.com/openzipkin/zipkin/pull/3561#issuecomment-1873495338) here. I'll create scaffolding in zipkin-otel so that it is easy to re-raise this PR and carry things forward.

PR #214, opened 2023-12-11
Updates all test dependencies including CI and removes Retrolambda

codefromthecrypt:
This updates all aspects of the build including GitHub Actions versions, maven versions and test dependency versions.  Most notably, this allows use of Java 11,17,21. Main versions are left for a different change.

To achieve this, we replace parts of the build that used RetroLambda to use maven profiles (similar to zipkin), as RetroLambda is no longer maintained. We also have to keep Java 11 in releases as later versions cannot compiler Java 1.6 bytecode anymore.


anuraaga (approved):
Does removing retrolambda being ok mean there weren't any lambdas in this repo?

codefromthecrypt:
> Does removing retrolambda being ok mean there weren't any lambdas in this repo?

In many cases, the lambda part of retrolambda didn't work. I looked and there was an out-of-date comment to that affect I deleted. The main value we were getting was rewriting the bytecode after the fact.

PR #216, opened 2023-12-11
deps: updates to latest libthrift

codefromthecrypt:
This updates to the most recent libthrift while bumping to the next minor. This is a compromised decision that takes into account that libthrift can cause revlocks in code that uses it, due in part to being perpetually <1.0 versions.

Fixes #201

PR #217, opened 2023-12-11
Stops publishing kafka08 sender

codefromthecrypt:
This stops publishing the kafka08 sender, beginning at the next minor version, 2.17. This has been unsupported by Kafka since 2015 and corresponding configuration removed in Zipkin in mid 2019. Removing this removes tech debt around the similarly abandoned junit rule used to test it.

See https://github.com/openzipkin/zipkin/pull/2564


codefromthecrypt:
depends on #216 which increments the minor version

codefromthecrypt:
rebased.

p.s. In the miniscule chance this is still needed someone can raise an issue to restore it, though this never happened for zipkin's repo where it was removed.

anuraaga (approved):

codefromthecrypt:
thanks for the looks @anuraaga!

PR #220, opened 2023-12-13
ci: moves ActiveMQ to docker based tests

codefromthecrypt:
this removes a junit 4 dependency as well

### DIFF
--- a/activemq-client/src/test/java/zipkin2/reporter/activemq/ActiveMQExtension.java
+++ b/activemq-client/src/test/java/zipkin2/reporter/activemq/ActiveMQExtension.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2016-2023 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.reporter.activemq;
+
+import java.time.Duration;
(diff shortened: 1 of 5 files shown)


activemq-client/src/test/java/zipkin2/reporter/activemq/ActiveMQExtension.java
```diff
@@ -0,0 +1,76 @@
...
+    ActiveMQConnectionFactory connectionFactory = new ActiveMQConnectionFactory();
+    connectionFactory.setBrokerURL(brokerURL());
+    return ActiveMQSender.newBuilder().queue(queue).connectionFactory(connectionFactory);
+  }
+
+  String brokerURL() {
+    return "failover:tcp://" + container.getHost() + ":" + container.getMappedPort(ACTIVEMQ_PORT);
+  }
+
+  // mostly waiting for https://github.com/testcontainers/testcontainers-java/issues/3537
```
anuraaga, line 64: I think this may be copy-pasted and no need to change anything now, but just for reference, `@TestContainers` and `@Container` automatically skip when docker isn't available. It would look something like this, but don't think it's a priority to change anything

Rename `ActiveMQExtension` -> `ActiveMQContainer`

```java
public static class ActiveMQContainer extends ... {
  public ActiveMQContainer() {
    ...
  }

  public String newSenderBuilder(String queue) {
    ...
  }
}

@Testcontainers
public class ITActiveMQSender {
  @Container public static final activemq = new ActiveMQContainer();

...
}
```

codefromthecrypt: does this also allow property override? in the main zipkin build we intentionally skip docker on one pass so we can parallelize them...

codefromthecrypt: I think if we could do what you said and still have a skip property (even if we make it), sounds much better and I can help sweep around and apply the pattern everywhere.

codefromthecrypt: I'll try a separate PR once this is in to do all of them. thanks for the tip @anuraaga!

PR #222, opened 2023-12-14
refactor: migrates from JUnit 4 to JUnit 5 (Jupiter)

codefromthecrypt:
This migrates from JUnit 4 to JUnit 5 (Jupiter) beginning with the following OpenRewrite command:

```bash
$ ./mvnw -U org.openrewrite.maven:rewrite-maven-plugin:run \
  -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-testing-frameworks:LATEST \
  -Drewrite.activeRecipes=org.openrewrite.java.testing.junit5.JUnit4to5Migration
```

After this, we had to do some work, notably as Brave needs to support non-alpha versions of OkHttp and alpha 5 creates a classpath conflict on Okio. Notably, I had to revert changes for the okhttp and urlconnection senders, which formerly used normal mockwebserver 3 and were automatically migrated to v5 alpha.

Next, I had to polish formatting as it undid some of that, and also fixed some cases where we were inconsistent. Then, I reviewed our base pom and removed some special cases that were no longer needed. Finally, I swept through and reduced visibility modifiers where possible.

Note: Both testcontainers and mockwebserver v3 have an implicit dep on junit 4. This is ok and the same as zipkin. The main thing here is we aren't explicitly using it anymore.

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -121,11 +121,6 @@
       <version>4.2.10</version>
     </dependency>
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>${junit.version}</version>
-    </dependency>
-    <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
       <version>${logback.version}</version>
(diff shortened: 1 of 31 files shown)


codefromthecrypt:
cc @timtebeek I think this particular OpenRewrite migration will also be rough in the other projects for reasons above, but yeah still helpful and hopefully paves a path for cleaner refactors later.

benchmarks/pom.xml
```diff
@@ -120,11 +120,6 @@
        https://github.com/charithe/kafka-junit -->
       <version>4.2.10</version>
```
codefromthecrypt, line 121: note we have kafka-junit here in the benchmark project, but probably we can migrate that to testcontainers at a low priority

anuraaga (approved):

codefromthecrypt:
thanks for letting me interrupt you yet another day in a row @anuraaga!

PR #223, opened 2023-12-14
okhttp3: pins main version to 4.x and tests 3.x in invoker

codefromthecrypt:
See https://github.com/openzipkin/zipkin-reporter-java/pull/205#issuecomment-1854178649 for discussion

### DIFF
--- a/okhttp3/pom.xml
+++ b/okhttp3/pom.xml
@@ -31,6 +31,8 @@
     <module.name>zipkin2.reporter.okhttp3</module.name>
 
     <main.basedir>${project.basedir}/..</main.basedir>
+    <!-- pre-1.8, pre-kotlin, and last version before Call.timeout. -->
+    <old-okhttp.version>3.11.0</old-okhttp.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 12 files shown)


okhttp3/pom.xml
```diff
@@ -31,6 +31,8 @@
     <module.name>zipkin2.reporter.okhttp3</module.name>
 
     <main.basedir>${project.basedir}/..</main.basedir>
+    <!-- pre-1.8, pre-kotlin, and last version before Call.timeout. -->
```
codefromthecrypt, line 34: from brave, as is most infrastructure in this PR, as brave does exactly the same thing for instrumentation. https://github.com/openzipkin/brave/blob/master/instrumentation/okhttp3/pom.xml#L34

anuraaga (approved):

pom.xml
```diff
@@ -63,9 +63,8 @@
     <brave.version>5.16.0</brave.version>
 
     <junit-jupiter.version>5.10.1</junit-jupiter.version>
-    <!-- use pre-kotlin version -->
-    <okhttp.version>3.14.9</okhttp.version>
-    <okhttp5.version>5.0.0-alpha.11</okhttp5.version>
```
codefromthecrypt, line 68: this wasn't referenced anywhere

PR #224, opened 2023-12-14
updates integration tests and benchmarks to use log4j

codefromthecrypt:
Similar to Brave, we don't want a large amount of clutter. To reduce a lot of it, I moved anything that uses MockWebServer to an IT (because it spams logs). I didn't move normal tests to log4j because we have a log capturing test that would fail. If someone wants to take a pass later at that, they could.


okhttp3/src/it/okhttp3_v3/pom.xml
```diff
@@ -68,26 +68,54 @@
...
         <version>@maven-surefire-plugin.version@</version>
         <configuration>
-          <!-- Ensures root cause ends up in the console -->
-          <trimStackTrace>false</trimStackTrace>
+          <failIfNoTests>true</failIfNoTests>
+          <includes>
+            <include>**/IT*.java</include>
+          </includes>
+          <!-- Try to prevent flakes in CI -->
+          <reuseForks>false</reuseForks>
```
codefromthecrypt, line 113: moved config from our normal failsafe

codefromthecrypt:
after this is in, I'll add invoker tests for the rabbitmq amqp-client which proves we can use recent versions and avoid CVE distraction while still maintaining compat (similar to okhttp). Finally, I'll do the same for spring and we're ready to roll ideally with no CVE scares.

anuraaga (approved):

codefromthecrypt:
Thanks for helping me consume so much Azure VM time @anuraaga!

PR #225, opened 2023-12-14
amqp-client: moves to latest version and tests prior

codefromthecrypt:
This follows the same pattern as OkHttp by testing RabbitMQ's prior version instead of pinning an old version with CVEs as a strict dependency. This removes the amount of CVE distractions and still keeps things compatible.


codefromthecrypt:
draft until #224

PR #232, opened 2023-12-18
deps: updates to zipkin 2.25.2

codefromthecrypt:
More important is the docker test image alignment, but also this helps applications pull less different recent versions of zipkin jar.


codefromthecrypt:
redundant #229

PR #234, opened 2024-01-05
Decouples from io.zipkin.zipkin2:zipkin

### LINKED ISSUE #182: zipkin-reporter-brave implementation has a sneaky dependency on zipkin2
I noticed that zipkin-reporter-brave still has a dependency on io.zipkin.zipkin2:zipkin because AsyncReporter and Sender extend zipkin2.Component among other things.

AsyncReporter
```java
import zipkin2.Call;
import zipkin2.CheckResult;
import zipkin2.Component;
import zipkin2.Span;
import zipkin2.codec.BytesEncoder;
import zipkin2.codec.SpanBytesEncoder;
```

Sender
```java
import zipkin2.Call;
import zipkin2.Component;
import zipkin2.codec.BytesEncoder;
import zipkin2.codec.Encoding;

```

We can't avoid a runtime dependency on io.zipkin.zipkin2:zipkin until some alternative surfaces. One way could be to extract simpler generic types that supports both brave and the old call sites without the Call abstraction.

codefromthecrypt:
This copies types from zipkin, so that reporter can be used without a dependency on it. In practice, many still will because encoders will be present. However, those using Brave and only JSON or alternative backends like Wavefront can now avoid a dependency on Zipkin.

Fixes #182
See https://github.com/micrometer-metrics/tracing/issues/509


reta:
Sorry for silly question @codefromthecrypt , I understood the purpose of the change but the overall picture is not entirely clear to me: there would be 2 sets of largely similar classes `zipkin2.*` and `zipkin2.reporter.*`. When `zipkin2.*` ones will be used? May be we could extract a new module (`zipkin2-reporter` fe) instead?

spring-beans/pom.xml
```diff
@@ -54,6 +54,12 @@
       <artifactId>zipkin-reporter</artifactId>
       <version>${project.version}</version>
```
reta, line 55: Exclusion here (as in `brave/pom.xml`)?

```
 <exclusions>
        <exclusion>
          <groupId>io.zipkin.zipkin2</groupId>
          <artifactId>zipkin</artifactId>
        </exclusion>
      </exclusions>
```

codefromthecrypt:
@reta good question. Actually, the amount of types reproduced were not big. In hindsight, we shouldn't have shared any base types here. I noticed this in #182. The generic type parameter on zipkin2.Reporter was thrwarted by an unremovable dependency on zipkin types. A couple years ago we made a more pliable MutableSpan type in brave, which has its own json encoder, so doesn't need the zipkin types.

So, the changes here are:

* For users of zipkin-reporter dep directly, only indirect types are hosted here, and likely will cause zero compilation breaks, as end users don't access things like Component or HealthCheckStatus (latter is internally used in the reporter).
* For users of zipkin-reporter-brave and json.. they have no dependency on io.zipkin anymore. This not only removes 200kb of deps for those only reporting in json, but also allows users like WaveFront and Stackdriver, who don't use zipkin models to avoid them too.
* For users of zipkin-reporter-brave and proto. You would need to either add the zipkin-reporter dep directly (which implicitly gets the zipkin types), or add the zipkin dep. That is until we add a direct proto encoder here, which someone could ask for.

The main point is the second, as I am finishing an update that will remove deprecated methods on Brave, and completely decouple it from zipkin types. Doing that and not also here would still drag them in.

codefromthecrypt:
Another side point is that when we do this decoupling, it is easier to eventually remove the floor Java 6 situation on zipkin's core jar. While this doesn't remove the problem entirely, it does for almost all Brave users, which is a large percentage of uses of zipkin-reporter

codefromthecrypt:
I am testing with a branch of both brave now (still zipkin incidental zipkin dep, but only for deprecated functions) and brave 6 (zipkin free due to deprecation removal). I'll test both with [brave-example](https://github.com/openzipkin/brave-example), then ready to merge if nothing breaks in those two.

spring-beans/pom.xml
codefromthecrypt: good eye, I was about 5 minutes from finding this out myself ;)

core/src/main/java/zipkin2/reporter/AsyncReporter.java
```diff
@@ -50,7 +33,8 @@
  *
  * @param <S> type of the span, usually {@link zipkin2.Span}
  */
-public abstract class AsyncReporter<S> extends Component implements Reporter<S>, Flushable {
+public final class AsyncReporter<S> extends Component implements Reporter<S>, Flushable {
```
anuraaga, line 36: I think going from abstract to final here might be a breaking change (not sure if you're intending for those yet). Abstract to non-final probably isn't if it's well behaved in terms of random subclasses

core/src/main/java/zipkin2/reporter/Reporter.java
```diff
@@ -24,16 +22,16 @@
  * zipkin2.Span).
  */
 public interface Reporter<S> {
-  Reporter<Span> NOOP = new Reporter<Span>() {
-    @Override public void report(Span span) {
+  Reporter<zipkin2.Span> NOOP = new Reporter<zipkin2.Span>() {
```
anuraaga, line 25: The change in this file appears to be a no-op, not sure if it's intended. Import statements are syntax sugar but don't think they have any effect on bytecode

core/src/main/java/zipkin2/reporter/AsyncReporter.java
codefromthecrypt: Because the constructor is package-private, I think it won't have any impact, right?

codefromthecrypt: I mean no one could have sub-classed it without sharing our package which wouldn't be supported anyway.

core/src/main/java/zipkin2/reporter/Reporter.java
codefromthecrypt: this was a relic from when I removed the dep entirely, which broke too much. reverted!

core/src/main/java/zipkin2/reporter/AsyncReporter.java
codefromthecrypt: I can take final off, but would want to make a comment. Would a reason be something like in case someone is doing bytecode instrumentation?

activemq-client/src/main/java/zipkin2/reporter/activemq/ActiveMQSender.java
```diff
@@ -19,13 +19,13 @@
 import javax.jms.JMSException;
 import javax.jms.QueueSender;
 import org.apache.activemq.ActiveMQConnectionFactory;
-import zipkin2.Call;
-import zipkin2.Callback;
-import zipkin2.CheckResult;
-import zipkin2.codec.Encoding;
 import zipkin2.reporter.AsyncReporter;
```
anuraaga, line 22: IIUC this import is only used in javadoc. For similar lines, just a nit but you may want to inline it instead to make it easier to scan the imports for usage of the zipkin-tied types, or lack of

codefromthecrypt: good idea. I'll look around for this!

core/src/main/java/zipkin2/reporter/AsyncReporter.java
codefromthecrypt: made a comment anyway!

codefromthecrypt:
I did some more dependency sweeping mostly about Brave 6, not this, and nearly ready to commit to this. I still need to test brave examples both with this on 5.18 and also 6.0. Then, we'll cut this dep and it to brave 5.18 and cut it, then also release Brave 6.0. Getting closer now!

codefromthecrypt:
It is hard to test this properly with jitpack. I will merge this and use sonatype snapshot repo to test with brave before releasing. Will still mention here the status.

codefromthecrypt:
Things are working in expected ways, so cutting 3.0. Thanks for all the help here!

---

This might be a good comment to bookmark @reta @shakuzen as basically brave-example tests run an end-to-end including running a trace through. If the docker test passes, basically it works.

Before
```bash
$ docker run --entrypoint /bin/sh --rm ghcr.io/openzipkin/brave-example:webmvc25-jetty -c 'find . -name \*jar|sort'
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
./root/WEB-INF/lib/aopalliance-1.0.jar
./root/WEB-INF/lib/brave-5.17.1.jar
./root/WEB-INF/lib/brave-context-log4j12-5.17.1.jar
./root/WEB-INF/lib/brave-instrumentation-http-5.17.1.jar
./root/WEB-INF/lib/brave-instrumentation-httpclient-5.17.1.jar
./root/WEB-INF/lib/brave-instrumentation-messaging-5.17.1.jar
./root/WEB-INF/lib/brave-instrumentation-rpc-5.17.1.jar
./root/WEB-INF/lib/brave-instrumentation-servlet-5.17.1.jar
./root/WEB-INF/lib/brave-instrumentation-spring-webmvc-5.17.1.jar
./root/WEB-INF/lib/brave-spring-beans-5.17.1.jar
./root/WEB-INF/lib/commons-codec-1.6.jar
./root/WEB-INF/lib/commons-logging-1.1.1.jar
./root/WEB-INF/lib/httpclient-4.3.6.jar
./root/WEB-INF/lib/httpcore-4.3.3.jar
./root/WEB-INF/lib/log4j-1.2.17.jar
./root/WEB-INF/lib/spring-beans-2.5.6.jar
./root/WEB-INF/lib/spring-context-2.5.6.jar
./root/WEB-INF/lib/spring-context-support-2.5.6.jar
./root/WEB-INF/lib/spring-core-2.5.6.jar
./root/WEB-INF/lib/spring-web-2.5.6.jar
./root/WEB-INF/lib/spring-webmvc-2.5.6.jar
./root/WEB-INF/lib/zipkin-2.27.0.jar
./root/WEB-INF/lib/zipkin-reporter-2.17.2.jar
./root/WEB-INF/lib/zipkin-reporter-brave-2.17.2.jar
./root/WEB-INF/lib/zipkin-reporter-spring-beans-2.17.2.jar
./root/WEB-INF/lib/zipkin-sender-urlconnection-2.17.2.jar
```

Now (no zipkin-N.N.N.jar)
```
$ ./build-bin/test webmvc25-jetty
$ docker run --entrypoint /bin/sh --rm openzipkin/brave-example:webmvc25-jetty-test -c 'find . -name \*jar|sort'
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
./root/WEB-INF/lib/aopalliance-1.0.jar
./root/WEB-INF/lib/brave-5.18.0-SNAPSHOT.jar
./root/WEB-INF/lib/brave-context-log4j12-5.18.0-SNAPSHOT.jar
./root/WEB-INF/lib/brave-instrumentation-http-5.18.0-SNAPSHOT.jar
./root/WEB-INF/lib/brave-instrumentation-httpclient-5.18.0-SNAPSHOT.jar
./root/WEB-INF/lib/brave-instrumentation-messaging-5.18.0-SNAPSHOT.jar
./root/WEB-INF/lib/brave-instrumentation-rpc-5.18.0-SNAPSHOT.jar
./root/WEB-INF/lib/brave-instrumentation-servlet-5.18.0-SNAPSHOT.jar
./root/WEB-INF/lib/brave-instrumentation-spring-webmvc-5.18.0-SNAPSHOT.jar
./root/WEB-INF/lib/brave-spring-beans-5.18.0-SNAPSHOT.jar
./root/WEB-INF/lib/commons-codec-1.6.jar
./root/WEB-INF/lib/commons-logging-1.1.1.jar
./root/WEB-INF/lib/httpclient-4.3.6.jar
./root/WEB-INF/lib/httpcore-4.3.3.jar
./root/WEB-INF/lib/log4j-1.2.17.jar
./root/WEB-INF/lib/spring-beans-2.5.6.jar
./root/WEB-INF/lib/spring-context-2.5.6.jar
./root/WEB-INF/lib/spring-context-support-2.5.6.jar
./root/WEB-INF/lib/spring-core-2.5.6.jar
./root/WEB-INF/lib/spring-web-2.5.6.jar
./root/WEB-INF/lib/spring-webmvc-2.5.6.jar
./root/WEB-INF/lib/zipkin-reporter-3.0.0-SNAPSHOT.jar
./root/WEB-INF/lib/zipkin-reporter-brave-3.0.0-SNAPSHOT.jar
./root/WEB-INF/lib/zipkin-reporter-spring-beans-3.0.0-SNAPSHOT.jar
./root/WEB-INF/lib/zipkin-sender-urlconnection-3.0.0-SNAPSHOT.jar
```

PR #236, opened 2024-01-06
build: fixes bytecode level and updates build versions

### LINKED ISSUE #235: Incorrect bytecode versions
## Describe the Bug

Using brave, I noticed the bytecode  versions aren't assigned correctly. For example, the URLConnectionSender should be compiled java 6 and it isn't.

```
 backend         | org.springframework.beans.factory.CannotLoadBeanClassException: Error loading class [zipkin2.reporter.beans.URLConnectionSenderFactoryBean] for bean with name 'sender' defined in ServletContext resource [/WEB-INF/applicationContext.xml]: problem with class file or dependent class; nested exception is java.lang.UnsupportedClassVersionError: zipkin2/reporter/beans/URLConnectionSenderFactoryBean : Unsupported major.minor version 52.0
```

## Steps to Reproduce

look here https://github.com/openzipkin/brave-example/pull/97

## Expected Behaviour

We probably need to lift infrastructure from zipkin to correct this.

codefromthecrypt:
This fixes a problem where we weren't releasing jars with the right bytecode level, due to a constant release=8. As we no longer can rely on RetroLambda to rewrite code, I've had to port things to the appropriate bytecode level manually. I also updated build deps that could be.

Fixes #235 and after this will cut a patch release (without any other pending PR)

### DIFF
--- a/core/src/test/java/zipkin2/reporter/AwaitableCallbackTest.java
+++ b/core/src/test/java/zipkin2/reporter/AwaitableCallbackTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 The OpenZipkin Authors
+ * Copyright 2016-2024 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -23,19 +23,14 @@ import static org.assertj.core.api.Assertions.failBecauseExceptionWasNotThrown;
 
 class AwaitableCallbackTest {
   @Test void awaitIsUninterruptable() {
-    AtomicBoolean returned = new AtomicBoolean();
     AwaitableCallback captor = new AwaitableCallback();
-    Thread thread = new Thread(() -> {
-      captor.await();
-      returned.set(true);
-    });
+    Thread thread = new Thread(captor::await);
     thread.start();
     thread.interrupt();
 
     assertThat(thread.isInterrupted()).isTrue();
     // The callback thread receiving an interrupt has nothing to do with the caller of the captor
     assertThat(Thread.currentThread().isInterrupted()).isFalse();
-    assertThat(returned.get()).isFalse();
   }
 
   @Test void onSuccessReturns() {
--- a/okhttp3/src/main/java/zipkin2/reporter/okhttp3/Platform.java
+++ b/okhttp3/src/main/java/zipkin2/reporter/okhttp3/Platform.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 The OpenZipkin Authors
+ * Copyright 2016-2024 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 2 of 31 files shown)


pom.xml
```diff
@@ -44,38 +44,41 @@
...
-    <main.java.version>1.8</main.java.version>
-    <main.signature.artifact>java18</main.signature.artifact>
-
-    <!-- default bytecode version for src/test -->
+    <!-- Except for the module named zipkin, all source is Java 8 -->
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
+    <!-- We don't use animal-sniffer anymore as release obviates it.
+         See https://github.com/mojohaus/animal-sniffer/issues/62 -->
+    <maven.compiler.release>8</maven.compiler.release>
```
codefromthecrypt, line 52: the problem was that we never overrode this, so the bytecode ended up as 8 on all projects

okhttp3/src/main/java/zipkin2/reporter/okhttp3/Platform.java
```diff
@@ -35,18 +35,31 @@ static Platform get() {
   static Platform findPlatform() {
     // Find JRE 8 new types
     try {
-      Class.forName("java.io.UncheckedIOException");
-      return new Jre8(); // intentionally doesn't not access the type prior to the above guard
+      Class<?> clazz = Class.forName("java.io.UncheckedIOException");
+      Constructor<?> ctor = clazz.getConstructor(IOException.class);
```
codefromthecrypt, line 39: this is annoying, but using this path is also very unlikely. It meant there was an IOException gzipping spans, which would be uncommon for sure.

core/src/test/java/zipkin2/reporter/AwaitableCallbackTest.java
```diff
@@ -23,19 +23,14 @@
...
-      returned.set(true);
-    });
+    Thread thread = new Thread(captor::await);
     thread.start();
     thread.interrupt();
 
     assertThat(thread.isInterrupted()).isTrue();
     // The callback thread receiving an interrupt has nothing to do with the caller of the captor
     assertThat(Thread.currentThread().isInterrupted()).isFalse();
-    assertThat(returned.get()).isFalse();
```
codefromthecrypt, line 38: this was winky depending on JDK. The important thing is that we interrupted the thread launched and not the current one, so I removed this part

codefromthecrypt:
cherry-picked a few classes and they look ok now

```bash
$ ./mvnw clean install  -Prelease -Dgpg.skip -Dmaven.javadoc.skip=true -DskipTests
--snip--
$ file core/target/classes//zipkin2/reporter/AsyncReporter.class
core/target/classes//zipkin2/reporter/AsyncReporter.class: compiled Java class data, version 50.0 (Java 1.6)
$ file brave/target/classes/zipkin2/reporter/brave/AsyncZipkinSpanHandler.class 
brave/target/classes/zipkin2/reporter/brave/AsyncZipkinSpanHandler.class: compiled Java class data, version 50.0 (Java 1.6)
$ file okhttp3/target/classes/zipkin2/reporter/okhttp3/OkHttpSender.class 
okhttp3/target/classes/zipkin2/reporter/okhttp3/OkHttpSender.class: compiled Java class data, version 51.0 (Java 1.7)
$ file spring-beans/target/classes/zipkin2/reporter/beans/URLConnectionSenderFactoryBean.class 
spring-beans/target/classes/zipkin2/reporter/beans/URLConnectionSenderFactoryBean.class: compiled Java class data, version 50.0 (Java 1.6)
```

PR #239, opened 2024-01-09
brave: allows AsyncZipkinSpanHandler to be build with custom formats

codefromthecrypt:
This is needed for stackdriver and any future OTLP work, as well as developing a standard proto encoder.


brave/src/main/java/zipkin2/reporter/brave/AsyncZipkinSpanHandler.java
```diff
@@ -151,18 +156,42 @@ public Builder queuedMaxBytes(int queuedMaxBytes) {
...
+    /**
+     * Builds an async span handler that encodes zipkin spans according to the encoder.
+     *
+     * <p>Note: The input encoder must use the same error tag implementation as configured by
+     * {@link #errorTag(Tag)}.
+     *
+     * @since 3.1
+     */
+    // AsyncZipkinSpanHandler not SpanHandler, so that Flushable and Closeable are accessible
+    public AsyncZipkinSpanHandler build(BytesEncoder<MutableSpan> encoder) {
```
codefromthecrypt, line 181: I noticed that we can't really move off zipkin types without this, as this handler only worked with json! Will release 3.1.0 shortly afterwards.

anuraaga (approved):

codefromthecrypt:
added tests and will merge on green. Thanks for the look @anuraaga!

codefromthecrypt:
after this https://github.com/openzipkin-contrib/zipkin-otel can have a brave encoder and sender without dependencies on zipkin types.

PR #241, opened 2024-01-10
ci: deploys bom separately to allow re-do on partial failure

codefromthecrypt:
Before, the bom project deployed only after the parent succeeded. This meant a timeout would prevent it from being deployed, even if the timeout was not critical. This deploys the bom independently so that such a situation doesn't cause us to have to re-cut a version.

This also fixes where we didn't actually use the release args in tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,4 +69,4 @@ jobs:
       # via forks, and login session ends up in ~/.docker. This is ok because
       # we publish DOCKER_PARENT_IMAGE to ghcr.io, hence local to the runner.
       - name: Test
-        run: build-bin/configure_test && build-bin/test
+        run: build-bin/configure_test && build-bin/test ${{ matrix.maven_args }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,7 +14,17 @@ on:
 
 jobs:
   deploy:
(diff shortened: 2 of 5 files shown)


codefromthecrypt:
this can be tested prior to release as it also runs on master

.github/workflows/test.yml
```diff
@@ -69,4 +69,4 @@ jobs:
       # via forks, and login session ends up in ~/.docker. This is ok because
       # we publish DOCKER_PARENT_IMAGE to ghcr.io, hence local to the runner.
       - name: Test
-        run: build-bin/configure_test && build-bin/test
+        run: build-bin/configure_test && build-bin/test ${{ matrix.maven_args }}
```
codefromthecrypt, line 72: @reta I feel like this copy/paste is missing elsewhere, too ;) that said I think the only ones that are critical on this are zipkin, reporter and brave, as they have java 6 artifacts..

codefromthecrypt:
will watch https://github.com/openzipkin/zipkin-reporter-java/actions/runs/7468624673 for success

codefromthecrypt:
worked.. will cut 3.1.1 and hope the sonatype snafu is over

.github/workflows/test.yml
reta: Got it, thanks @codefromthecrypt !

PR #244, opened 2024-01-13
Deprecates Sender for much simpler BytesMessageSender

codefromthecrypt:
Yesterday @shakuzen @anuraaga @kojilin and I chatted about how the async reporters never actually use the async code of the senders. The async side was only used by zipkin-server, as the async reporter design intentionally uses a blocking loop to flush the span backlog. So, this makes things simpler by only requiring a blocking implementation: `BytesMessageSender.send` (chosen to not create symbol collisions and similar to `BytesMessageEncoder`.

The side effect is that new senders can use this interface and completely avoid the complicated `Call` then `execute` chain, deferring to whatever the library-specific blocking path is. I have implemented this in armeria as a test, and it is absolutely easier.

As older spring libraries like sleuth may not upgrade, this maintains the old types until reporter 4. However, new senders can simply implement the new type and be done with it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Reporter.CONSOLE.report(span);
 
 ## AsyncReporter
 AsyncReporter is how you actually get spans to zipkin. By default, it waits up to a second
-before flushes any pending spans out of process via a Sender.
+before flushes any pending spans out of process via a BytesMessageSender.
 
 ```java
 reporter = AsyncReporter.create(URLConnectionSender.create("http://localhost:9411/api/v2/spans"));
@@ -67,7 +67,7 @@ Here are the most important properties to understand when tuning.
 Property | Description
 --- | ---
 `queuedMaxBytes` |  Maximum backlog of span bytes reported vs sent. Corresponds to `ReporterMetrics.updateQueuedBytes`. Default 1% of heap
-`messageMaxBytes` | Maximum bytes sendable per message including overhead. Default `500,000` bytes (`500KB`). Defined by `Sender.messageMaxBytes`
+`messageMaxBytes` | Maximum bytes sendable per message including overhead. Default `500,000` bytes (`500KB`). Defined by `BytesMessageSender.messageMaxBytes`
 `messageTimeout` |  Maximum time to wait for messageMaxBytes to accumulate before sending. Default 1 second
 `closeTimeout` |  Maximum time to block for in-flight spans to send on close. Default 1 second
 
@@ -84,11 +84,12 @@ by a large `messageTimeout` or `messageMaxBytes`. Consider lowering the
 `messageMaxBytes` if this occurs, as it will result in less work per
 message.
 
-## Sender
+## BytesMessageSender
 The sender component handles the last step of sending a list of encoded spans onto a transport.
-This involves I/O, so you can call `Sender.check()` to check its health on a given frequency.
+This involves I/O, so you can call `sender.send(Collections.emptyList())` to check it works before
+using.
 
-Sender is used by AsyncReporter, but you can also create your own if you need to.
+BytesMessageSender is used by AsyncReporter, but you can also create your own if you need to.
 ```java
 class CustomReporter implements Flushable {
 
@@ -99,7 +100,12 @@ class CustomReporter implements Flushable {
 
   // Is the connection healthy?
   public boolean ok() {
-    return sender.check().ok();
+    try {
+      sender.send(Collections.emptyList());
+      return true;
+    } catch (Exception e) {
+      return false;
+    }
   }
 
   public void report(Span span) {
@@ -113,7 +119,7 @@ class CustomReporter implements Flushable {
     pending.drainTo(drained);
     if (drained.isEmpty()) return;
 
-    sender.sendSpans(drained, callback).execute();
+    sender.send(drained);
   }
 ```
 
--- a/core/src/main/java/zipkin2/reporter/BytesMessageSender.java
+++ b/core/src/main/java/zipkin2/reporter/BytesMessageSender.java
@@ -109,7 +109,11 @@ public interface BytesMessageSender extends Closeable {
   /**
    * Sends a list of encoded spans to a transport such as HTTP or Kafka.
    *
-   * @param encodedSpans list of encoded spans.
+   * <p><em>Empty input is permitted</em>. While async reporters in this repository will always send
+   * a non-empty list. Some external callers might use an empty send for fail-fast checking. If you
+   * obviate empty lists, you might break them. See /RATIONALE.md for more.
+   *
+   * @param encodedSpans a potentially empty list of encoded spans.
    * @throws IllegalStateException if {@link #close() close} was called.
    */
   void send(List<byte[]> encodedSpans) throws IOException;
--- a/core/src/main/java/zipkin2/reporter/BytesMessageSender.java
+++ b/core/src/main/java/zipkin2/reporter/BytesMessageSender.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2016-2024 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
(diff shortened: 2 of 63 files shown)


codefromthecrypt:
PS this will also make things a lot easier for porting. For example, the spring boot senders are pretty much all synchronous anyway and so the code can be shredded once this is released. e.g. https://github.com/spring-projects/spring-boot/blob/7851c2362eb53c989cbdda706576c729a016bc49/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/HttpSender.java#L43

cc @mhalbritter

codefromthecrypt:
here's an example sender, for armeria, granted I didn't add gzip to it yet.. the lack of creating the `Call` objects makes it a lot easier as you'll see vs okhttp for example (not its fault, as it would also be simple if we could remove Call without deprecation here) https://github.com/openzipkin/zipkin-reporter-java/pull/245

README.md
```diff
@@ -84,11 +84,12 @@ by a large `messageTimeout` or `messageMaxBytes`. Consider lowering the
 `messageMaxBytes` if this occurs, as it will result in less work per
 message.
 
-## Sender
+## BytesMessageSender
 The sender component handles the last step of sending a list of encoded spans onto a transport.
-This involves I/O, so you can call `Sender.check()` to check its health on a given frequency.
```
codefromthecrypt, line 89: almost everything implemented this with a post to empty list, and in any case `check` was almost never called!

core/src/main/java/zipkin2/reporter/BytesMessageSender.java
```diff
@@ -0,0 +1,116 @@
...
+ * scribe will likely write each span as a separate log line.
+ *
+ * <p>This accepts a list of {@link BytesEncoder#encode(Object) encoded spans}, as opposed a list of
+ * spans like {@code zipkin2.Span}. This allows senders to be re-usable as model shapes change. This
+ * also allows them to use their most natural message type. For example, kafka would more naturally
+ * send messages as byte arrays.
+ *
+ * @since 3.2
+ */
+public interface BytesMessageSender extends Closeable {
```
codefromthecrypt, line 45: here's the type which defines only methods actually used

core/src/main/java/zipkin2/reporter/Sender.java
```diff
@@ -40,42 +41,13 @@
  * send messages as byte arrays.
  *
  * @since 3.0
+ * @deprecated since 3.2, use {@link BytesMessageSender} instead. This will be removed in v4.0.
  */
-public abstract class Sender extends Component {
+@Deprecated
+public abstract class Sender extends Component implements BytesMessageSender {
```
codefromthecrypt, line 47: retrofitted Sender as BytesMessageSender so that non-updated code can drop in and work. This is also why BytesMessageSender is an interface, as we can't extend two classes

core/src/main/java/zipkin2/reporter/internal/AsyncReporter.java
```diff
@@ -273,7 +278,7 @@ void flush(BufferNextMessage<S> bundler) {
       });
 
       try {
-        sender.sendSpans(nextMessage).execute();
```
codefromthecrypt, line 276: this was the only place where the sender was used for its real purpose in this repo. So, the net change here is just more efficient and straightforward of the same

core/src/main/java/zipkin2/reporter/BytesMessageSender.java
```diff
@@ -0,0 +1,116 @@
...
+   */
+  int messageSizeInBytes(int encodedSizeInBytes);
+
+  /**
+   * Sends a list of encoded spans to a transport such as HTTP or Kafka.
+   *
+   * @param encodedSpans list of encoded spans.
+   * @throws IllegalStateException if {@link #close() close} was called.
+   */
+  void send(List<byte[]> encodedSpans) throws IOException;
```
reta, line 115: (This is just an opinion) The semantics of sending an empty list as to way to very the connection could be surprising - logically I think noop is expected. May be having a dedicated check / ping method is not as bad after all

reta:
Looks pretty clean, but letting other folks (who were part of the discussion) to take a look

core/src/main/java/zipkin2/reporter/BytesMessageSender.java
codefromthecrypt: I really regret having put this check in the first place as it was a relic from finagle failfast. It isn't used for anything, but it is totally the case that sending an empty list is permitted today. I would rather document an empty list is permitted since almost all impls of check had to send an empty list. This is better than forcing people to implement check which would only do the same, and then people think it is smarter than that.

codefromthecrypt: please have a look at the new RATIONALE.md for notes on this!

reta: Thanks a lot @codefromthecrypt for very comprehensive clarification! (knowing context always pays off)

core/src/main/java/zipkin2/reporter/internal/AsyncReporter.java
```diff
@@ -42,17 +46,18 @@
...
+ * limited in speed to what a single thread can clear. When a thread cannot clear the backlog, new
+ * spans are dropped.
  *
  * @param <S> type of the span, usually {@code zipkin2.Span}
  * @since 3.0
  */
-public abstract class AsyncReporter<S> extends Component implements Reporter<S>, Flushable {
-  public static Builder newBuilder(Sender sender) {
+public abstract class AsyncReporter<S> extends Component implements Reporter<S>, Closeable, Flushable {
+  public static Builder newBuilder(BytesMessageSender sender) {
```
anuraaga, line 60: This is compile compatible but not sure if linkage compatible, it might be but forget the java details on this. Could be worth a double check

anuraaga (approved):

core/src/main/java/zipkin2/reporter/internal/AsyncReporter.java
codefromthecrypt: I'll put in a test to instantiate a deprecated sender. Good point.

codefromthecrypt: PTAL

anuraaga (approved):

codefromthecrypt:
Thanks for the thought @reta and @anuraaga!

codefromthecrypt:
https://github.com/openzipkin/brave-example/pull/104 shows this works, so I'll release it so people can play with it, also to help with the maintenance work ahead for the spring hosted senders.

codefromthecrypt:
Here's a working armeria sender in less than 50 lines including imports:

```java
package brave.example;

import com.linecorp.armeria.client.WebClient;
import com.linecorp.armeria.common.AggregatedHttpResponse;
import com.linecorp.armeria.common.HttpData;
import com.linecorp.armeria.common.HttpMethod;
import com.linecorp.armeria.common.HttpRequest;
import com.linecorp.armeria.common.MediaType;
import java.io.IOException;
import java.util.List;
import zipkin2.reporter.BytesMessageEncoder;
import zipkin2.reporter.BytesMessageSender;
import zipkin2.reporter.ClosedSenderException;
import zipkin2.reporter.Encoding;

final class WebClientSender extends BytesMessageSender.Base {
  final WebClient zipkinApiV2SpansClient;
  volatile boolean closeCalled; // volatile as not called from the reporting thread.

  WebClientSender(WebClient zipkinApiV2SpansClient) {
    super(Encoding.JSON);
    this.zipkinApiV2SpansClient = zipkinApiV2SpansClient;
  }

  @Override public int messageMaxBytes() {
    return 500_000; // Use the most common HTTP default
  }

  @Override public void send(List<byte[]> encodedSpans) throws IOException {
    if (closeCalled) throw new ClosedSenderException();
    byte[] body = BytesMessageEncoder.JSON.encode(encodedSpans);
    HttpRequest request =
        HttpRequest.of(HttpMethod.POST, "", MediaType.JSON, HttpData.wrap(body));
    AggregatedHttpResponse response = zipkinApiV2SpansClient.blocking().execute(request);
    try (HttpData content = response.content()) {
      if (!response.status().isSuccess()) {
        if (content.isEmpty()) {
          throw new IOException("response failed: " + response);
        }
        throw new IOException("response for failed: " + content.toStringAscii());
      }
    }
  }

  @Override public void close() {
    closeCalled = true;
  }
}
```

core/src/main/java/zipkin2/reporter/internal/AsyncReporter.java
codefromthecrypt: you were right and my tests didn't catch this. invoker could have but anyway https://github.com/openzipkin/zipkin-reporter-java/pull/246

PR #245, opened 2024-01-13
Adds Armeria WebClientSender

codefromthecrypt:
This adds an option to use Armeria as a sender. The first user will be zipkin-server and brave-example. cc @minwoox @anuraaga @kojilin

### DIFF
--- a/armeria/src/main/java/zipkin2/reporter/armeria/WebClientSender.java
+++ b/armeria/src/main/java/zipkin2/reporter/armeria/WebClientSender.java
@@ -1,0 +1,164 @@
+/*
+ * Copyright 2016-2024 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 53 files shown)


armeria/src/main/java/zipkin2/reporter/armeria/WebClientSender.java
```diff
@@ -0,0 +1,164 @@
...
+ * <p>This sender is thread-safe.
+ */
+public final class WebClientSender extends BytesMessageSender.Base {
+
+  /** Creates a sender that posts {@link Encoding#JSON} messages. */
+  public static WebClientSender create(String endpoint) {
+    return newBuilder(WebClient.of(endpoint)).build();
+  }
+
+  public static Builder newBuilder(WebClient client) {
```
codefromthecrypt, line 55: I actually wanted this so I could control the endpoint group with eureka

codefromthecrypt:
this or something like it might be better off in armeria with the other tracing stuff. I'll use a copy of it in brave-example meanwhile

codefromthecrypt:
closing as I think this should be in armeria itself, as it easier to manage versions. Also, one day a truly async brave span handler could be made there, should anyone want to.

PR #249, opened 2024-02-11
Adds HttpEndpointSupplier to allow generic discovery of HTTP endpoints

codefromthecrypt:
Before, users could only use spring-cloud-sleuth to integrate HTTP senders with dynamic endpoint suppliers such as consul, eureka or zookeeper. Since sleuth is discontinued, this adds a lightweight generic functionality via a new type `HttpEndpointSupplier` which can do the same sort of translation as sleuth did (e.g. substituting a placeholder URI for a real one).

This will allow those upgrading to Spring Boot 3 to continue using service discovery such as Eureka. As an additional benefit, it also allows the non-spring ecosystem the same.

There is no current plan to introduce Eureka or any other built-in endpoint discovery mechanics to this repo. However, tests will prove they can be integrated, and the types will be validated in config scenarios including Spring Boot and raw Armeria prior to merge. If users request specific discovery providers, they can raise issues and demand for them.

### DIFF
--- a/core/src/main/java/zipkin2/reporter/HttpEndpointSupplier.java
+++ b/core/src/main/java/zipkin2/reporter/HttpEndpointSupplier.java
@@ -1,3 +1,16 @@
+/*
+ * Copyright 2016-2024 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
 package zipkin2.reporter;
 
 import java.util.List;
--- a/core/src/main/java/zipkin2/reporter/HttpEndpointSupplier.java
+++ b/core/src/main/java/zipkin2/reporter/HttpEndpointSupplier.java
@@ -19,8 +19,20 @@ import java.util.List;
  * HTTP-based {@link BytesMessageSender senders} use this to resolve a potentially-pseudo endpoint
  * passed by configuration to a real endpoint.
  *
- * <p>Senders should consider the special value {@link #FIXED_FACTORY} and the type {@link Fixed} to
- * avoid dynamic lookups when constants will be returned.
+ * <h3>Usage Notes</h3>
+ *
+ * <p>Sender should implement the following logic:
+ * <ul>
+ *   <li>During build, the sender should invoke the {@linkplain Factory}.</li>
+ *   <li>If the result is {@link Fixed}, build the sender to use a static value.</li>
+ *   <li>Otherwise, call {@link HttpEndpointSupplier#get()} each time
+ *       {@linkplain BytesMessageSender#send(List)} is invoked.</li>
+ * </ul>
+ *
+ * <h3>Implementation Notes</h3>
+ *
+ * <p>Implement friendly {@code toString()} functions, that include the real endpoint or the one
+ * passed to the {@linkplain Factory}.
  *
  * <p>Senders are not called during production requests, rather in time or size bounded loop, in a
  * separate async reporting thread. Implementations that resolve endpoints via remote calls, such as
@@ -35,13 +47,13 @@ import java.util.List;
  */
 public interface HttpEndpointSupplier {
   /**
-   * HTTP {@link BytesMessageSender sender} builders check for this symbol, and will substitute its
-   * input as a fixed endpoint value rather than perform dynamic lookups.
+   * HTTP {@link BytesMessageSender sender} builders check for this symbol, and return the input as
+   * a {@linkplain Fixed} result rather than perform dynamic lookups.
    *
    * @since 3.3
    */
   Factory FIXED_FACTORY = new Factory() {
-    @Override public HttpEndpointSupplier create(String endpoint) {
+    @Override public Fixed create(String endpoint) {
       return new Fixed(endpoint);
     }
   };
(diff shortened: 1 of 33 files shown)


codefromthecrypt:
TODO: actually integrate

core/src/main/java/zipkin2/reporter/HttpEndpointSupplier.java
```diff
@@ -0,0 +1,84 @@
...
+ * passed by configuration to a real endpoint.
+ *
+ * <p>Senders should consider the special value {@link #FIXED_FACTORY} and the type {@link Fixed} to
+ * avoid dynamic lookups when constants will be returned.
+ *
+ * <p>Senders are not called during production requests, rather in time or size bounded loop, in a
+ * separate async reporting thread. Implementations that resolve endpoints via remote calls, such as
+ * from Eureka, should cache internally to avoid blocking the reporter thread on each loop.
+ *
+ * <p>Some senders, such as Armeria, may have more efficient and precise endpoint group logic. In
```
codefromthecrypt, line 16: For this part, I might put an example in the test tree. While this will pull armeria into our deps that are analyzed with snyk etc, usually there are no CVEs that aren't resolved quickly. Also the dependency tree isn't as huge as spring boot, so it isn't as bad for laptops in hotels.

core/src/main/java/zipkin2/reporter/HttpEndpointSupplier.java
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Factory passed to HTTP {@link BytesMessageSender sender} builders to control resolution of the
+   * static endpoint from configuration.
+   *
+   * <p>Unless this is {@linkplain #FIXED_FACTORY}, {@linkplain #create(String)} will be deferred to
+   * the first call to {@linkplain BytesMessageSender#send(List)}.
+   *
+   * @since 3.3
+   */
+  interface Factory {
```
codefromthecrypt, line 55: this type will be added to spring-beans so that anyone using spring can employ it via config. that's currently the only config layer we have in this repo.

codefromthecrypt:
ok done except 3rd-party integration tests which I'll report back in brave-example.

okhttp3/src/main/java/zipkin2/reporter/okhttp3/OkHttpSender.java
```diff
@@ -113,20 +116,30 @@ public static final class Builder {
...
+     * No default. The POST HttpUrl for zipkin's <a href="https://zipkin.io/zipkin-api/#/">v2 api</a>,
      * usually "http://zipkinhost:9411/api/v2/spans"
      */
     // customizable so that users can re-map /api/v2/spans ex for browser-originated traces
     public Builder endpoint(String endpoint) {
       if (endpoint == null) throw new NullPointerException("endpoint == null");
-      HttpUrl parsed = HttpUrl.parse(endpoint);
-      if (parsed == null) throw new IllegalArgumentException("invalid POST url: " + endpoint);
-      return endpoint(parsed);
+      this.endpoint = endpoint;
```
reta, line 136: I think the `endpoint` is not needed anymore, is it? (we could use `endpointSupplierFactory` solely):

```java
this.endpointSupplierFactory = (HttpEndpointSupplier.Constant) endpoint -> endpoint;
```

codefromthecrypt: I should note where we are assigning this, that we need to retain the old endpoint or else we break the former (toBuilder()) functions. will make a comment!

codefromthecrypt: for example, if someone changes like .. toBuilder().endpointSupplierFactory(foo).build() we don't want to erase the (config) endpoint used to create the supplier

codefromthecrypt: added comments and edge case tests!

reta: Got it, thanks @codefromthecrypt !

reta (approved):

codefromthecrypt:
thanks for the review @reta! since jitpack doesn't work anymore, I'm going to merge and address anything else post-commit. This significantly reduces work for integration because for example, to try this in micrometer requires their snapshots, too!

core/src/main/java/zipkin2/reporter/HttpEndpointSupplier.java
```diff
@@ -0,0 +1,115 @@
...
+ * @since 3.3
+ */
+public interface HttpEndpointSupplier extends Closeable {
+  /**
+   * HTTP {@link BytesMessageSender sender} builders check for this symbol, and return the input as
+   * a {@linkplain Constant} result rather than perform dynamic lookups.
+   *
+   * @since 3.3
+   */
+  Factory CONSTANT_FACTORY = new Factory() {
```
anuraaga, line 57: Just want to confirm this is fine to be public (private interface statics are post Java 8). Otherwise, a static method constructing a package private class can be less surface

core/src/main/java/zipkin2/reporter/HttpEndpointSupplier.java
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * HTTP {@link BytesMessageSender senders} check for this type, and will cache its first value.
+   *
+   * @since 3.3
+   */
+  final class Constant implements HttpEndpointSupplier {
+    private final String endpoint;
+
+    public Constant(String endpoint) {
```
anuraaga, line 103: Related to above, this looks like the "real" public API, a static method may encapsulate better

core/src/main/java/zipkin2/reporter/HttpEndpointSupplier.java
codefromthecrypt: yep I added this so there can be a default bean in spring without thinking too much (they replicate our senders)

codefromthecrypt: anyway no problem. just was following how like in brave we have constants like this, but it isn't important to continue that.

core/src/main/java/zipkin2/reporter/HttpEndpointSupplier.java
codefromthecrypt: ok I'll move this out to a different file and make the function. no problem.

codefromthecrypt: actually I can just hide the ctor, we still need the type public

codefromthecrypt: actually if I move things I have to move everything to a new type, and now the interface has to refer to things in another file to talk about behavior. this is unlike the rest of the codebase, but I'll raise a PR and you can decide.

codefromthecrypt:
Thanks for the review @anuraaga #251  for follow-up

PR #250, opened 2024-02-12
Decreases switch statements and deprecates BytesMessageEncoder

codefromthecrypt:
I noticed in our senders and reporters there were multiple internal `switch` statements to get functions for a specific encoding. This is worse externally for zipkin-reporter-brave, as call sites need to duplicate switch statements, and also update them in order to allow another encoding. This is the case with PROTO3, for example, which is not yet supported for brave spans.

Here are the main changes:

* Add `Encoding.encode(List<byte[]>)` from the now deprecated `BytesMessageEncoder`.
* Add `Encoding.mediaType()`, useful in internal and external HTTP senders.
* Re-use the existing zipkin `SpanBytesEncoder.forEncoding(Encoding)` to reduce redundant switch logic.
* Add `MutableSpanBytesEncoder.forEncoding(Encoding)` for Brave
* Add `MutableSpanBytesEncoder.create(Encoding, Tag<Throwable>)` for Brave users who want an alt error tag.

With these changes in place, external senders such as those defined in spring-boot, can inherit sensible defaults without maintenance.

This acknowledges some specialized senders (such as okhttp) won't use `Encoding.encode(List<byte[]>)`. However, this change makes commodity senders a lot easier to write and configure, as they can now be driven by just `Encoding` and an endpoint.

### DIFF
--- a/brave/src/main/java/zipkin2/reporter/brave/MutableSpanBytesEncoder.java
+++ b/brave/src/main/java/zipkin2/reporter/brave/MutableSpanBytesEncoder.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2016-2024 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.reporter.brave;
+
+import brave.Tag;
+import brave.Tags;
+import brave.handler.MutableSpan;
+import zipkin2.reporter.BytesEncoder;
+import zipkin2.reporter.Encoding;
+
+/** Includes built-in formats used in Zipkin. */
+public enum MutableSpanBytesEncoder implements BytesEncoder<MutableSpan> {
+  /** Corresponds to the Zipkin v2 json format */
+  JSON_V2 {
+    @Override public Encoding encoding() {
+      return Encoding.JSON;
+    }
+
+    @Override public int sizeInBytes(MutableSpan input) {
+      return JsonV2Encoder.INSTANCE.sizeInBytes(input);
+    }
+
+    @Override public byte[] encode(MutableSpan input) {
+      return JsonV2Encoder.INSTANCE.encode(input);
+    }
+  };
+
+  /**
+   * Returns the default {@linkplain MutableSpan} encoder for given encoding.
+   *
+   * @throws UnsupportedOperationException if the encoding is not yet supported.
+   * @since 3.3
+   */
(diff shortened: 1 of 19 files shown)


brave/src/main/java/zipkin2/reporter/brave/MutableSpanBytesEncoder.java
```diff
@@ -0,0 +1,81 @@
...
+   * @throws UnsupportedOperationException if the encoding is not yet supported.
+   * @since 3.3
+   */
+  public static BytesEncoder<MutableSpan> forEncoding(Encoding encoding) {
+    if (encoding == null) throw new NullPointerException("encoding == null");
+    switch (encoding) {
+      case JSON:
+        return JSON_V2;
+      case PROTO3:
+        throw new UnsupportedOperationException("PROTO3 is not yet a built-in encoder");
```
codefromthecrypt, line 52: cool thing is that I have a pending change to support this. However, anyone using `forEncoding` could update into it without prior knowledge.

codefromthecrypt:
seems like spring boot need to release soon, so I will check on this now. hopefully, looks ok to the team, but if you have a chance to take a look, please do!

anuraaga (approved):

codefromthecrypt:
Here are some notable diffs.

```diff 
-       HttpSender() {
-               super(Encoding.JSON);
+       HttpSender(Encoding encoding) {
+               super(encoding);
        }
 
        @Override
@@ -74,7 +74,7 @@ abstract class HttpSender extends BytesMessageSender.Base {
                if (this.closed) {
                        throw new ClosedSenderException();
                }
-               byte[] body = BytesMessageEncoder.JSON.encode(encodedSpans);
+               byte[] body = this.encoding.encode(encodedSpans);
                HttpHeaders headers = getDefaultHeaders();
                if (needsCompression(body)) {
                        body = compress(body);
@@ -86,7 +86,7 @@ abstract class HttpSender extends BytesMessageSender.Base {
        HttpHeaders getDefaultHeaders() {
                HttpHeaders headers = new HttpHeaders();
                headers.set("b3", "0");
-               headers.set("Content-Type", "application/json");
+               headers.set("Content-Type", this.encoding.mediaType());
                return headers;
        }
```

```diff
                @Bean
                @ConditionalOnMissingBean(value = MutableSpan.class, parameterizedContainer = BytesEncoder.class)
                @ConditionalOnEnabledTracing
-               BytesEncoder<MutableSpan> braveSpanEncoder() {
-                       return new JsonV2Encoder();
+               BytesEncoder<MutableSpan> braveSpanEncoder(Encoding encoding) {
+                       return MutableSpanBytesEncoder.forEncoding(encoding);
                }
 
                @Bean
@@ -157,8 +158,8 @@ class ZipkinConfigurations {
                @Bean
                @ConditionalOnMissingBean(value = Span.class, parameterizedContainer = BytesEncoder.class)
                @ConditionalOnEnabledTracing
-               BytesEncoder<Span> zipkinSpanEncoder() {
-                       return SpanBytesEncoder.JSON_V2;
+               BytesEncoder<Span> zipkinSpanEncoder(Encoding encoding) {
+                       return SpanBytesEncoder.forEncoding(encoding);
                }
```

reta (approved):

codefromthecrypt:
thanks for the look @anuraaga @reta! I need to add the long overdue brave proto encoder next to avoid folks running into a wall. It is a somewhat mechanical port of what's in zipkin and brave already..

PR #253, opened 2024-02-13
Adds HttpSender to aid proper handling of HttpEndpointSupplier

codefromthecrypt:
I noticed that implementing constant and dynamic endpoint handing properly is a chore, especially to test it. This adds `HttpSender` as a second base type. It handles the edge cases so implementors don't have to.

I also added `SenderAdapter` which helps deal with the Call contract stuff until we can delete it. I also deleted some async tests from okhttp as we don't promise that sort of behavior on a deprecated and unused function. (For example, otel no longer use the async call stuff).


core/src/test/java/zipkin2/reporter/HttpSenderTest.java
```diff
@@ -0,0 +1,187 @@
...
+import zipkin2.Span;
+import zipkin2.codec.SpanBytesEncoder;
+
+import static java.util.stream.Collectors.toList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static zipkin2.TestObjects.CLIENT_SPAN;
+
+@ExtendWith(MockitoExtension.class)
+class HttpSenderTest {
```
codefromthecrypt, line 36: tests here were duplicated between urlconnection and okhttp senders. since the behavior is proven here, it is less work to port the resttemplate and webclient senders in boot, as I don't need to copy/paste these tests and burden folks to review them

codefromthecrypt:
have one more change on top of this that I thought of overnight, but unrelated, so will make another PR

PR #254, opened 2024-02-14
Makes constant endpoint an api with implementations available

codefromthecrypt:
In a former change @anuraaga rightly mentioned that there is trouble putting concrete types in the HttpSupplier interface file. I noticed spring has its own type they use as a constant supplier, `ZipkinConnectionDetails`. Since this is an interface, this double-underscores the same concern.

This *hopefully* clears up the topic, as well making a place to implement a rate-limited supplier, something sleuth has, but boot doesn't yet.


anuraaga (approved):

reta (approved):

reta:
Looks pretty clean! 🥇

codefromthecrypt:
glad you like it, missing rate limiter coming in a minute. This is important as eureka isn't designed to respond fast (at least I notice some latency). We also need it to rate limit when there are errors.

PR #257, opened 2024-03-05
license: removes copyright year and uses SPDX ID

codefromthecrypt:
As a small project, we have to conserve resources and not sign up for work that isn't required. I've recently realized many [commercial](https://github.com/hashicorp/consul/blob/main/command/cli/cli.go#L1-L2) and/or [CNCF projects](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/Tracer.java#L1-L4) both use SPDX IDs and also don't bother with copyright year. If their legal team is ok with this, surely a volunteer team without access to one should be, too!

Doing so accomplishes the following, all of which help volunteers:
* significantly increases readability of files, particularly small ones
* removes beginning of year maintenance, which cause a lot of FUD last year. IIRC some deployment failed and hours were spent in spite of docs.
* eliminates the need to do a full source check out, just to satisfy the license plugin. This means we can use actions/checkout defaults.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -32,7 +32,6 @@ jobs:
           # Prevent use of implicit GitHub Actions read-only token GITHUB_TOKEN.
           # We push Javadocs to the gh-pages branch on commit.
           token: ${{ secrets.GH_TOKEN }}
-          fetch-depth: 0  # allow build-bin/idl_to_gh_pages to get the full history
       - name: Setup java
         uses: actions/setup-java@v4
         with:
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -44,31 +44,6 @@ Here's an example of a snapshot deploy with specified credentials.
 $ export GPG_TTY=$(tty) && GPG_PASSPHRASE=whackamole SONATYPE_USER=adrianmole SONATYPE_PASSWORD=ed6f20bde9123bbb2312b221 build-bin/build-bin/maven/maven_deploy
 ```
 
-## First release of the year
-
-The license plugin verifies license headers of files include a copyright notice indicating the years a file was affected.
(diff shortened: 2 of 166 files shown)


.github/workflows/deploy.yml
```diff
@@ -32,7 +32,6 @@ jobs:
           # Prevent use of implicit GitHub Actions read-only token GITHUB_TOKEN.
           # We push Javadocs to the gh-pages branch on commit.
           token: ${{ secrets.GH_TOKEN }}
-          fetch-depth: 0  # allow build-bin/idl_to_gh_pages to get the full history
```
anuraaga, line 35: Didn't check the code just doublechecking this is fine the comment isn't about "license check"

RELEASE.md
```diff
@@ -44,31 +44,6 @@ Here's an example of a snapshot deploy with specified credentials.
 $ export GPG_TTY=$(tty) && GPG_PASSPHRASE=whackamole SONATYPE_USER=adrianmole SONATYPE_PASSWORD=ed6f20bde9123bbb2312b221 build-bin/build-bin/maven/maven_deploy
 ```
 
-## First release of the year
```
anuraaga, line 47: 💪 💪

anuraaga (approved):

.github/workflows/deploy.yml
codefromthecrypt: I should have self-commented. this repo doesn't have a script like this, so it is fuzz

anuraaga: Well I guess it's copy pasta from a different repo :D

codefromthecrypt: yeah the dreaded zipkin-api repo ;)

codefromthecrypt:
thanks for advice and the review @anuraaga

PR #263, opened 2024-04-12
ci: adds lint check, fixes yaml formatting and unnecessary auth

codefromthecrypt:
This adds new scripts configure_lint and lint which work like our other build-bin scripts. These lint the workflow yaml and our markdown links.

I also removed the separate step for javadoc checking as it is easy to do that in the release test target, as well uses less resources vs doing it twice.

See https://github.com/openzipkin/zipkin-gcp/pull/212 for the first impl of this.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -40,7 +40,7 @@ jobs:
         with:
           path: ~/.m2/repository
           key: ${{ runner.os }}-jdk-11-maven-${{ hashFiles('**/pom.xml') }}
-          restore-keys: ${{ runner.os }}-jdk-17-maven-
+          restore-keys: ${{ runner.os }}-jdk-11-maven-
       - name: Deploy
         env:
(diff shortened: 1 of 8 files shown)


.github/workflows/deploy.yml
```diff
@@ -22,42 +23,39 @@ jobs:
...
+          # Deploy the Bill of Materials (BOM) separately as it is unhooked
+          # from the main project intentionally.
           - name: bom
             deploy_script: build-bin/deploy_bom
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
-        with:
-          # Prevent use of implicit GitHub Actions read-only token GITHUB_TOKEN.
-          # We push Javadocs to the gh-pages branch on commit.
```
codefromthecrypt, line 33: this was a lie, as in this repo we don't!

reta (approved):

codefromthecrypt:
thanks for the eye @reta!

PR #267, opened 2024-04-14
Adds SECURITY.md and scanning workflow

codefromthecrypt:
This adds SECURITY.md and a scanning workflow, using Trivy. In particular, this clarifies what we use to scan for vulnerabilities (Trivy, not anything else), and the only channel likely to be responded to on a significant issue (zipkin-admin email, not advisories as people ignored them).

I won't merge this until I get at least 2 approves. After that, I'll help apply this to the other java repos.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,0 +1,52 @@
+---
+name: security
+
+# We don't scan documentation-only commits.
+on:  # yamllint disable-line rule:truthy
+  push:  # non-tagged pushes to master
+    branches:
+      - master
+    tags-ignore:
+      - '*'
+    paths-ignore:
+      - '**/*.md'
+      - './build-bin/*lint'
+      - ./build-bin/mlc_config.json
+  pull_request:  # pull requests targeted at the master branch.
+    branches:
+      - master
+    paths-ignore:
+      - '**/*.md'
+      - './build-bin/*lint'
+      - ./build-bin/mlc_config.json
+
+jobs:
+  security:
+    name: Security
+    runs-on: ubuntu-22.04  # newest available distribution, aka jellyfish
+    # skip commits made by the release plugin
+    if: "!contains(github.event.head_commit.message, 'maven-release-plugin')"
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - uses: actions/cache@v4
+        name: Cache Trivy Database
+        with:
+          path: .trivy
+          key: ${{ runner.os }}-trivy
+          restore-keys: ${{ runner.os }}-trivy
+      - name: Run Trivy vulnerability and secret scanner
+        uses: aquasecurity/trivy-action@master
+        id: trivy
+        with:
+          scan-type: 'fs'
+          scan-ref: '.'  # scan the entire repository
+          scanners: vuln,secret
+          exit-code: '1'
+          severity: HIGH,CRITICAL
+          output: trivy-report.md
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
ps all the main repos are clean, with the exception of zipkin-dependencies, which is a mess due to documented revlock on spark. When it gets to that repo, I'll make a special case note in the workflow.

.github/workflows/security.yml
```diff
@@ -0,0 +1,52 @@
...
+        id: trivy
+        with:
+          scan-type: 'fs'
+          scan-ref: '.'  # scan the entire repository
+          scanners: vuln,secret
+          exit-code: '1'
+          severity: HIGH,CRITICAL
+          output: trivy-report.md
+          cache-dir: .trivy
+      - name: Set Summary
```
codefromthecrypt, line 49: intentionally just setting the summary vs opting into GitHub features. When we run trivy like this, it is basically the same as `trivy fs .` so quite easy and has no licensing or other impact with GitHub, nor any new tabs to accidentally ignore.

reta (approved):

SECURITY.md
reta, line 1: 👍

codefromthecrypt:
@anuraaga @reta @llinder @shakuzen trivya:

The reason why this doesn't show up in red, despite testing with old spring deps, is that Trivy agreed that [maven-invoker-test deps should be classified as dev dependencies](https://github.com/aquasecurity/trivy/discussions/5787), and they don't trigger by default anymore.

That we could resolve that issue in OSS transparently, and also that trivy is OSS and we can run manually like all our other linters, is the reason I went with trivy here. If you think this decision should be in RATIONALE.md, I'm happy to add it!

reta:
> If you think this decision should be in RATIONALE.md, I'm happy to add it!

I think it is all good, not need for `RATIONALE.md`, thanks @codefromthecrypt !

.github/workflows/security.yml
```diff
@@ -0,0 +1,52 @@
...
+        shell: bash
+        if: ${{ failure() && steps.trivy.conclusion == 'failure' }}
+        run: cat trivy-report.md >> $GITHUB_STEP_SUMMARY
```
anuraaga, line 52: Might be worth adding a link or comment about the 1MB limit on step summary.

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#step-isolation-and-limits

anuraaga (approved):

.github/workflows/security.yml
codefromthecrypt: hopefully we don't end up with over 1MB of CVE descriptions, but sure ;)

codefromthecrypt: put a note with a cheeky suggestion if this happens :D

codefromthecrypt:
Thanks for the feedback @reta @anuraaga. As promised, I'll run this around the maintained repos in the org so that it isn't "half done"

codefromthecrypt:
PS I spent a good bit of time getting zipkin-dependencies clean, so this means all our prod images can use the same security settings.

It is likely zipkin-dependencies will need more maintenance than others to stay clean. I think as volunteers we can choose how often we want to do that work.

https://github.com/openzipkin/zipkin-dependencies/pull/226

codefromthecrypt:
OK I raised PRs for all the most maintained and/or production repos. Folks who are interested can raise them for any of the others.

PR #270, opened 2024-09-09
Avoid encoding list of spans twice in BaseHttpSender

shakuzen:
newBody was called twice which would result in encoding the list of spans twice unnecessarily. This change will reduce allocations and copying memory.


codefromthecrypt (approved):
great catch!

codefromthecrypt:
@reta wanna do one of your version bumps prior to cutting a patch release?

reta:
> @reta wanna do one of your version bumps prior to cutting a patch release?

Sure! Here we are: https://github.com/openzipkin/zipkin-reporter-java/pull/271

PR #271, opened 2024-09-09
Routine dependency updates

reta:
Routine dependency updates

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
 
     <!-- override to set exclusions per-project -->
     <errorprone.args />
-    <errorprone.version>2.26.1</errorprone.version>
+    <errorprone.version>2.30.0</errorprone.version>
 
     <!-- use the same value in bom/pom.xml -->
     <!-- NOTE: Do not update to 3.x as that has floor Java 8 -->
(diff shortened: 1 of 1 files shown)


shakuzen (approved):

codefromthecrypt (approved):
thanks again! wanna do the honors of pushing the release tag?

shakuzen:
It looks like deploying the bom was successful but the jars failed: https://github.com/openzipkin/zipkin-reporter-java/actions/runs/10792310306
Can we re-run the job or do we need to manually do something to finish the release?

codefromthecrypt:
smells like a sonatype timeout. maybe double-check they didn't make it to central. If not, double check with the sonatype oss login. regardless, running again shouldn't hurt as maven prevents double-uploading server-side

p.s. I noticed they changed since I last looked to require an email instead of a username login https://ossindex.sonatype.org/user/signin

shakuzen:
It looks like none of the jars made it to central ([dependabot PR to upgrade](https://app.circleci.com/pipelines/github/micrometer-metrics/tracing/1716/workflows/ca88015a-1d14-40a9-ac12-2fa81884eb27/jobs/8153) is failing to find any of the jars). I've re-run the deploy jars job now and [it was successful](https://github.com/openzipkin/zipkin-reporter-java/actions/runs/10792310306/job/29968516175). Things should be in central soon.

PR #277, opened 2025-02-12
Fix bug when send multi messages in pulsar sender

CodePrometheus:
- Fix bug when send multi messages in pulsar sender
- Add corresponding test

### DIFF
--- a/pulsar-client/src/test/java/zipkin2/reporter/pulsar/ITPulsarSender.java
+++ b/pulsar-client/src/test/java/zipkin2/reporter/pulsar/ITPulsarSender.java
@@ -5,6 +5,7 @@
 package zipkin2.reporter.pulsar;
 
 import org.apache.pulsar.client.api.Consumer;
+import org.apache.pulsar.client.api.Message;
 import org.apache.pulsar.client.api.SubscriptionInitialPosition;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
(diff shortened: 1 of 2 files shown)


pulsar-client/src/test/java/zipkin2/reporter/pulsar/ITPulsarSender.java
```diff
@@ -97,6 +98,30 @@ class ITPulsarSender {
...
+      int size = 10;
+      for (int i = 0; i < size; i++) {
+        send(sender, CLIENT_SPAN);
+      }
+      for (int i = 0; i < size; i++) {
+        assertThat(SpanBytesDecoder.JSON_V2.decodeList(readMessage(sender)))
+          .hasSize(1).containsExactly(CLIENT_SPAN);
+      }
+      assertThatThrownBy(() -> SpanBytesDecoder.JSON_V2.decodeList(readMessage(sender)))
+        .isInstanceOf(NullPointerException.class);
```
shakuzen, line 114: A NPE is a bit mysterious of a result for there being no messages to read. Also, the test takes at least 10 seconds now as it waits for the 10 second timeout on the `consumer.receive` call. It'd be ideal (I think) for the test to only take such time when it is failing.

shakuzen (approved):
Thanks for the quick fix. The test shows this works now. I left a comment for consideration about the test but no need to block merging on it.

pulsar-client/src/test/java/zipkin2/reporter/pulsar/ITPulsarSender.java
CodePrometheus: After reconsidering, this doesn’t add much here, let me polish this, maybe better to just remove it.

